PR #3, opened 2022-12-20
Stubbing out a few more endpoints

dcramps:
Some stubs for a few things for the login process. Very WIP, merge if you want to continue on it, otherwise I'll pick it up again tomorrow.

- `ut/api/cloudstorage/user/{id}/` endpoints. These handle user settings and progression. Need to handle GET and PUT. Store in Mongo somehow (byte array?)
- `entitlement/api/account/{id}/entitlements` returns nothing, should return the static data.
- `oauth/verify` endpoint just returns OK. Need to determine if that's all it needs. Data [here](https://github.com/MixV2/EpicResearch/blob/master/docs/account/endpoints/verify_auth_session.md) is not complete.
- `QueryProfile`/`mmrbulk`/`league` endpoints all returning nothing since matchmaking is gonna be dead / last priority
- `AccountDataService` intended for getting the binary profile/progression data. Unimplemented.


timiimit (approved):
# Verify
I assume that `oauth/verify` endpoint also extends expiration of token without changing the token itself, otherwise game would stop working after some hours of being open and user being logged in. but i could be wrong.

# Rating/matchmaking
i'm pretty sure that `mmrbulk` needs to be implemented, it returns ratings of the player, which is displayed in scoreboards. for now we can just make everyone have default 1500 rating in all modes.
example request:
```json
{
	"ratingTypes": [
		"SkillRating",
		"TDMSkillRating",
		"DMSkillRating",
		"CTFSkillRating",
		"ShowdownSkillRating",
		"FlagRunSkillRating",
		"RankedDuelSkillRating",
		"RankedCTFSkillRating",
		"RankedShowdownSkillRating",
		"RankedFlagRunSkillRating"
	]
}
```
example response:
```json
{
	"ratingTypes":[
		"SkillRating",
		"TDMSkillRating",
		"DMSkillRating",
		"CTFSkillRating",
		"ShowdownSkillRating",
		"FlagRunSkillRating",
		"RankedDuelSkillRating",
		"RankedCTFSkillRating",
		"RankedShowdownSkillRating",
		"RankedFlagRunSkillRating"
	],
	"ratings":[1655,1862,2038,1827,1626,2630,1666,1500,1500,1500],
	"numGamesPlayed":[546,1029,2073,181,17,2264,29,0,0,0]
}
```
Perhaps we can make qp put you in normal hub matches.

# CloudStorage
~~I would rename `AccountDataService` to `CloudStorageService` since that's what it seems you want it to handle.~~

I think we should support cloudstorage as a whole, but just `user_profile_2` and `user_progression_1` files would suffice for now and should probably be stored in a database. Everyone has these two files.

Editor is capable of uploading pak files when trying to share content. i'll try to figure out the protocol. cloudstorage is basically a file hosting service like https://ut4pugs.us/redirect or https://utcc.unrealpugs.com that no one never knew existed. Perhaps if this server project takes off, we can connect it with one or both of those services and host files there. that way content creators would be able to directly publish content from editor for everyone to see.

PR #5, opened 2022-12-20
Full cloudstorage support

timiimit:
# Description
This PR adds full cloudstorage support. db accepts only two files that everyone has: `user_progression_1` and `user_profile_2`. each file needs to be smaller than 1MB. mongodb supposedly has limit of 16MB, but since both of those files are way smaller i just limited their size to 1MB.

systemfiles need to be added to db before hand. they can be found at `<game_install_location>\UnrealTournament\PersistentDownloadDir\EMS`. systemfiles are identified in db with `AccountID == EpicID.Empty`,

# Missing work
There is one thing missing and that is to identify the correct hashing algorithms that epic uses when returning a list of files.

### DIFF
--- a/UT4MasterServerDotNetCore/Controllers/CloudStorageController.cs
+++ b/UT4MasterServerDotNetCore/Controllers/CloudStorageController.cs
@@ -1,49 +1,107 @@
 ﻿using Microsoft.AspNetCore.Mvc;
+using Newtonsoft.Json.Linq;
 using UT4MasterServer.Authorization;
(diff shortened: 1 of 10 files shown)


timiimit:
Alright, i just checked if i guessed the hashing algorithms correctly and i did. `hash` in response is SHA1 and `hash256` is SHA256.

timiimit:
I missed one thing... listing systemfiles should also add `"doNotCache": false` json to returned objects

PR #19, opened 2022-12-29
Verify that `EpicID` is a hexadecimal string

Saibamen:


timiimit:
i already did this earlier today

Saibamen:
@timiimit: But you created custom code. Why not to use standard C# methods?

timiimit:
Are you sure your code would work? That method is used to parse an int, which is 4 bytes long. the id string parameter can represent a 16 byte number.

Saibamen:
> Are you sure your code would work? That method is used to parse an int, which is 4 bytes long. the id string parameter can represent a 16 byte number.

I wrote xUnit tests and your code is better: https://github.com/timiimit/UT4MasterServer/pull/32

PR #23, opened 2022-12-29
Docker build

Saibamen:
Build Docker image on every commit and Pull Request

Example: https://github.com/Saibamen/UT4MasterServer/actions/runs/3800915657


timiimit:
I know nothing about github actions or docker but ill trust that this is useful

PR #27, opened 2022-12-29
Frontend: SHA512 and basic validation

Saibamen:


timiimit:
i'm not sure if we want to set a minimum of 7 chars for username. the game seems to work with just 1 character. maybe 3 or 4 should be minimum. this is currently just a client side requirement which i don't see as a problem.

i haven't decided on email requirement yet. i think it could be optional, and only set by users who want a more secure account by allowing them to recover account via email. I'm even not sure if sending emails is feasible for this project.

someone is already working on a separate server for frontend which will make api calls to this one (tho i havent seen any progress from them for over 1 week), so there is no need to work extensively on these minimal frontend pages. i just intend them to be used for development purposes.

Saibamen:
> i'm not sure if we want to set a minimum of 7 chars for username

For password...

For Username, Epic has requirement from 3 to 16 characters

timiimit:
My bad, I misread that.
all good then

PR #32, opened 2022-12-29
xUnit tests for IsHexString

Saibamen:
See: https://github.com/timiimit/UT4MasterServer/pull/19#issuecomment-1367379242

### DIFF
--- a/UT4MasterServer/Other/EpicID.cs
+++ b/UT4MasterServer/Other/EpicID.cs
@@ -1,13 +1,12 @@
 ﻿using MongoDB.Bson.Serialization;
 using MongoDB.Bson.Serialization.Attributes;
-using System.Text.Json.Serialization;
 
 namespace UT4MasterServer;
 
 [Serializable]
 public struct EpicID : IComparable<EpicID>, IEquatable<EpicID>, IConvertible, IBsonClassMapAttribute
 {
-	private readonly static Random r = new Random();
+	private static readonly Random r = new Random();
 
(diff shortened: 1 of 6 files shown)


UT4MasterServer/Other/EpicID.cs
```diff
@@ -28,11 +27,30 @@ private EpicID(string id)
...
+	}
+
+	public static bool IsHexString(string input)
+	{
+		if (string.IsNullOrWhiteSpace(input))
+		{
+			return false;
+		}
+
+		input = input.ToLower();
```
timiimit, line 43: Now `string.ToLower` is called twice for no reason when `EpicID.IsHexString` is called from constructor.
I think there should be an additional condition for uppercase characters in the foreach loop below instead of converting whole string to lower *just* to have one condition less.

timiimit (changes requested):

UT4MasterServer/Other/EpicID.cs
```diff
@@ -28,11 +27,30 @@ private EpicID(string id)
...
 		ID = id.ToLower();
 
-		foreach (char c in ID)
+		if (!IsHexString(ID))
+		{
+			throw new ArgumentException("id needs to be a hexadecimal string");
+		}
+	}
+
+	public static bool IsHexString(string input)
```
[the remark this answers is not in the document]
Saibamen: Done

UT4MasterServer/Other/EpicID.cs
Saibamen: Done

timiimit (approved):
Looks good

PR #37, opened 2022-12-30
Parsing request header

DannyAllegrezza:
chore:  Make sure incoming request value is formatted correctly
    This is just a tiny quality of life improvement. I personally believe
    the server should be able to handle whether a client sends `BeArEr xxx`
    or `bearer xxx`

    This will also make it easier to send requests using postman or swagger

### DIFF
--- a/UT4MasterServer/Authentication/HttpAuthorization.cs
+++ b/UT4MasterServer/Authentication/HttpAuthorization.cs
@@ -7,8 +7,8 @@ public class HttpAuthorization
 	public string Type { get; set; } = string.Empty;
 	public string Value { get; set; } = string.Empty;
 
-	public bool IsBearer { get => Type == "bearer"; }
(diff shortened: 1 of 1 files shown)


timiimit:
If the main point is to improve development environment, i think this should be in `#if DEBUG` `#endif` block

Saibamen:
More work was done here: https://github.com/timiimit/UT4MasterServer/pull/38

DannyAllegrezza:
I'm used to working with the .NET framework where it handles parsing out the value from the request headers when the `Authorization: Bearer` strategy is used. AFAIK, the framework doesn't care about the case sensitivity of the word "bearer" as the value of a header key. So I was honestly just trying to match that behavior without refactoring the entire approach right now. 

I would say for a future improvement, we could look at taking more advantage of the built in Identity library!

Also, I guess we can close this as #38 covers making changes to the `HttpAuthorization` class.

PR #38, opened 2022-12-30
Add auth button to Swagger. Fix namespaces

Saibamen:
image

### DIFF
--- a/UT4MasterServer/Authentication/HttpAuthorization.cs
+++ b/UT4MasterServer/Authentication/HttpAuthorization.cs
@@ -1,16 +1,20 @@
 ﻿using System.Net;
 
-namespace UT4MasterServer.Authorization;
+namespace UT4MasterServer.Authentication;
 
 public class HttpAuthorization
 {
 	public string Type { get; set; } = string.Empty;
 	public string Value { get; set; } = string.Empty;
 
-	public bool IsBearer { get => Type == "bearer"; }
-	public bool IsBasic { get => Type == "basic"; }
+	public const string AuthorizationHeader = "Authorization";
+	public const string BearerScheme = "Bearer";
+	public const string BasicScheme = "Basic";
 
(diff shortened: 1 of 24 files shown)


timiimit (approved):

UT4MasterServer/Authentication/HttpAuthorization.cs
```diff
@@ -1,16 +1,20 @@
...
 
 public class HttpAuthorization
 {
 	public string Type { get; set; } = string.Empty;
 	public string Value { get; set; } = string.Empty;
 
-	public bool IsBearer { get => Type == "bearer"; }
-	public bool IsBasic { get => Type == "basic"; }
+	public const string AuthorizationHeader = "Authorization";
+	public const string BearerScheme = "Bearer";
```
DannyAllegrezza, line 11: I believe a decent portion of the behavior of the Authentication/Authorization middleware we've written may not be necessary. There is a nice built in middleware for JWT based auth that I'd suggest we consider using. Unless of course there is a reason why we need to things on our own?

The docs overview are here; https://learn.microsoft.com/en-us/aspnet/core/security/authentication

A short explanation of JWT stuff here: https://learn.microsoft.com/en-us/aspnet/core/signalr/authn-and-authz?view=aspnetcore-6.0#built-in-jwt-authentication

It isn't a huge deal, but the `Microsoft.AspNetCore.Authentication.JwtBearer` package includes const values out of the box, so whether we write the code ourselves or pull it in from an official package is really up to us. For example, you could use the `JwtBearerDefaults.AuthenticationScheme` which is the same const string of "Bearer".

image

Saibamen: Yes, I know about this `Microsoft.AspNetCore.Authentication.JwtBearer`

I just added Auth button to Swagger and I didn't want to rewrite our auth middleware without testing it (I didn't changed my local `Engine.ini` yet :D). Adding NuGet package here to only reference 2 constants was not necessary IMO 

If you want to change anything, create new PR

PR #40, opened 2022-12-30
Initial commit, mostly stubs

caseyswilliams:
Stubbed out a web front end, log in is almost ready to be functional, but I haven't gotten the API running in vscode yet to test. Will need to make some docker compose changes to get both sites running easily.
image


caseyswilliams:
Note: I created this as a separate site as to not hinder current functionality while I finish it out. I will work on getting it containerized and into compose so the development experience will be straightforward.

timiimit (approved):
I don't have much js/ts experience but from quick glance over ir. i Approve.

Saibamen:
Are you sure you want to use Vue to only few pages?

Why not using just Bootstrap for login and register page?

React is smaller and doesn't force you to follow requirements for big web apps

UT4MasterServer.Web/src/services/authentication.service.ts
```diff
@@ -0,0 +1,37 @@
+import { ILoginRequest } from '../types/login-request';
+
+export default class AuthenticationService {
+    private tokenUrl = `${__UT4UU_API_URL}/account/api/oauth/token`;
+    private authUrl = `${__UT4UU_API_URL}/account/api/oauth/auth`;
```
Saibamen, line 5: Why using `__UT4UU_API_URL` instead of proxy?

caseyswilliams: These are environment variables for storing the api url for different environments, if they exist. Not strictly required. A proxy is certainly possible in vite, but i dont see why it would be necessary as the web app and api are served separately. As for vue/ts, I always prefer strong typing over weak, and I just like vue better than react. As you can see there really isn't much in the way of boilerplate. By default, react is actually slightly larger than vue (100kb vs 80kb for vue)

caseyswilliams: Oh, and I am using bootstrap

caseyswilliams: As for whether I'm sure, I'm sure it's my preference, this only took me about 3 hours to do this much. It will only be about another 6 to finish, possibly a little more to get the docker compose setup for easy development. I chose the frameworks I'm most comfortable with because that is what I'll be fastest at and we're all just doing this in our free time. But if there are to be other contributors to this portion and it is a problem I'd be happy to let someone else do the setup and stubs and I'll finish them up/help out. Whatever helps get it across the finish line.

PR #41, opened 2022-12-30
Throw EPIC errors in some controllers + fix some responses

Saibamen:
Add more TODOs
`killType.ToUpper()`

Tested with Swagger on https://account-public-service-prod03.ol.epicgames.com/

### DIFF
--- a/UT4MasterServer/Controllers/AccountController.cs
+++ b/UT4MasterServer/Controllers/AccountController.cs
@@ -2,10 +2,14 @@
 using Microsoft.AspNetCore.Mvc;
 using Newtonsoft.Json.Linq;
 using UT4MasterServer.Authentication;
+using UT4MasterServer.Models;
 using UT4MasterServer.Services;
 
 namespace UT4MasterServer.Controllers;
 
+/// <summary>
+/// account-public-service-prod03.ol.epicgames.com
+/// </summary>
 [ApiController]
 [Route("account/api")]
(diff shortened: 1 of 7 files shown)


UT4MasterServer/Controllers/AccountController.cs
```diff
@@ -67,6 +80,19 @@ public async Task<IActionResult> GetAccounts([FromQuery(Name = "accountId")] Lis
 		if (User.Identity is not EpicUserIdentity authenticatedUser)
 			return Unauthorized();
 
+		if (accountIDs.Count == 0 || accountIDs.Count > 100)
+		{
+			return NotFound(new ErrorResponse
+			{
+				ErrorCode = "errors.com.epicgames.account.invalid_account_id_count",
+				ErrorMessage = "Sorry, the number of account id should be at least one and not more than 100.",
```
timiimit, line 88: Is this an actual response message from epic server?
If not, then i think there should be plural `account ids`.

timiimit (changes requested):
Nice! You got an extensive amount of errors and permissions. I honestly didn't even expect us to handle those.

There's just one thing im unsure about. I left a code comment.

UT4MasterServer/Controllers/AccountController.cs
Saibamen: This is from EPIC prod

PR #43, opened 2023-01-03
`LeagueTier` names

Saibamen:
From https://github.com/EpicGames/UnrealTournament/blob/clean-master/UnrealTournament/Source/UnrealTournament/Private/UTPlayerState.cpp#L2219

### DIFF
--- a/UT4MasterServer/Models/League.cs
+++ b/UT4MasterServer/Models/League.cs
@@ -3,12 +3,17 @@ using System.Text.Json.Serialization;
 
 namespace UT4MasterServer.Models;
 
+/// <summary>
+/// Names taken from <see href="https://github.com/EpicGames/UnrealTournament/blob/clean-master/UnrealTournament/Source/UnrealTournament/Private/UTPlayerState.cpp#L2219">LeagueTierToText</see>
+/// </summary>
 public enum LeagueTier
 {
-	None = 0, // idk how this is named, its when you don't have colored badge on stock scoreboard
-	Bronze = 1,
-	Silver = 2,
-	Gold = 3
+	BronzeLeague = 0,
+	SilverLeague = 1,
+	GoldLeague = 2,
+	PlatinumLeague = 3,
+	MasterLeague = 4,
+	GrandMasterLeague = 5,
 }
 
 public class League
@@ -18,7 +23,7 @@ public class League
 	// TODO: Use one JSON DLL in solution. See https://github.com/timiimit/UT4MasterServer/issues/33
 	[JsonPropertyName("tier")] // Fix for API response
 	[JsonProperty("tier")]
-	public LeagueTier Tier { get; set; } = LeagueTier.None;
+	public LeagueTier Tier { get; set; } = LeagueTier.BronzeLeague;
 
 	[JsonPropertyName("division")]
 	[JsonProperty("division")]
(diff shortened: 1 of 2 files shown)


UT4MasterServer/Authentication/BasicAuthenticationHandler.cs
```diff
@@ -26,6 +27,11 @@ protected override async Task<AuthenticateResult> HandleAuthenticateAsync()
 				return AuthenticateResult.Fail("unexpected scheme");
 			}
 
+			if (!authorization.Value.IsBase64())
+			{
+				return AuthenticateResult.Fail("not Base64");
+			}
+
```
timiimit, line 34: I would prefer if this was combined inside `ClientIdentification` class. it's constructor already tries to decode base64...

UT4MasterServer/Helpers/StringExtensions.cs
```diff
@@ -19,4 +19,15 @@ public static bool IsHexString(this string input)
...
+
+	public static bool IsBase64(this string input)
+	{
+		if (string.IsNullOrWhiteSpace(input))
+		{
+			return false;
+		}
+
+		var buffer = new Span<byte>(new byte[input.Length]);
+		return Convert.TryFromBase64String(input, buffer, out _);
```
timiimit, line 31: Instead of just checking `IsBase64` we could **additionally** expose `string.TryDecodeBase64` with an `output` parameter for decoded string.

timiimit (changes requested):

UT4MasterServer/Controllers/SessionController.cs
```diff
@@ -148,7 +148,10 @@ public class SessionController : JsonAPIController
 
 		if (account == null)
 			account = await accountService.GetAccountAsync(session.AccountID);
-		logger.LogInformation($"User {account} was authorized via {grantType}");
+		/* TODO: Account still can be null if session.AccountID is empty GUID, for example, if we are using access_token from client_credentials grant type.
+		 * What to do here?
+		 */
+		logger.LogInformation($"User '{account}' was authorized via {grantType}");
```
timiimit, line 154: I don't think there's an issue with `account` being `null`. If it's null we just don't print user related info in response. Json generating code below does account null checking.

For the sake of logging there should be `account ?? EpicID.Empty.ToString()` in logged string.

UT4MasterServer/Helpers/StringExtensions.cs
Saibamen: Done

UT4MasterServer/Authentication/BasicAuthenticationHandler.cs
Saibamen: Done

UT4MasterServer/Controllers/SessionController.cs
Saibamen: It gives me `Operator '??' cannot be applied to operands of type 'UT4MasterSever.Models.Account?' and 'UT4MasterServer.Other.EpicID'?`, so I just deleted TODO

timiimit: Then i suppose instead of  `account ?? EpicID.Empty` use `account?.ToString() ?? EpicID.Empty.ToString()`

Saibamen: Done

PR #49, opened 2023-01-03
Return `ErrorResponse` when invalid credentials in password grant_type

Saibamen:
And return `ErrorResponse` when password is not allowed + with message about using UT4UU

## Before:

image

## After:

image

### DIFF
--- a/UT4MasterServer/Controllers/SessionController.cs
+++ b/UT4MasterServer/Controllers/SessionController.cs
@@ -109,17 +109,7 @@ public class SessionController : JsonAPIController
 				//       really need multi-factor auth. it is after all the way that ut's login screen
 				//       works when you start the game without launcher (and without UT4UU).
 
-				if (username == null)
-				{
-					return ErrorInvalidRequest("username");
-				}
-
-				if (password == null)
-				{
-					return ErrorInvalidRequest("password");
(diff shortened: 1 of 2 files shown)


UT4MasterServer/Controllers/SessionController.cs
```diff
@@ -122,10 +119,34 @@ public class SessionController : JsonAPIController
 					return ErrorInvalidRequest("password");
 				}
 
-				// TODO: Check permission and return ErrorResponse: Sorry your client is not allowed to use the grant type password. errorCode: errors.com.epicgames.common.oauth.unauthorized_client
+				// Yeah, EPIC is returning this ErrorResponse after checking username and password...
+				if (!allowPasswordGrant)
+				{
+					return BadRequest(new ErrorResponse
```
timiimit, line 125: I understand that epic is checking credentials first, but I don't think this is desired. Even though its just null checks before it, I would prefer if this was returned prior to any checks at location where `if (!allowPasswordGrant)` was before.

Maybe make a comment saying we are doing things differently than epic.

Saibamen: Done

timiimit (approved):

PR #51, opened 2023-01-05
Fix Authenticate method

timiimit:
This fixes an issue with `Authenticate` method. We need to do what game expects.


UT4MasterServer/Models/Requests/AuthenticateRequest.cs
```diff
@@ -1,22 +1,24 @@
-﻿using System.Text.Json.Serialization;
+﻿using Microsoft.AspNetCore.Mvc;
+using System.Text.Json.Serialization;
```
Saibamen, line 2: Unused using I think

Issue #22, opened 2022-12-29
Handle statistics


timiimit:
## Description
Statistics need to be saved in the database. These include among other things, number of kills per each weapon, number of deaths per each weapon, hits with certain weapon damage type, etc.

### How stats work
endpoint `/ut/api/stats/accountId/{id}/bulk/window/{statsType}` is used to retrieve all stats. and `stats.json` in cloudstorage contains score of last 5 played games.

All statistics are collected via analytics. that is on domain `datarouter.ol.epicgames.com` and endpoint `/datarouter/api/v1/public/data` + query that looks in my case like this:
```
?SessionID=%7BD9873B18-401E-A352-9120-D3BB8FB2E4CB%7D&AppID=UnrealTournament.Dev&AppVersion=1.0.0.0%20-%20%2B%2BUT%2BMain-CL-3525109&UserID=9c808754479eaa36e7efe896f204da11%7C0b0f09b400854b9b98932dd9e5abe7c5%7C3d42b762-3e87-4d29-90d8-7f36f619c3ae%7C%7COSS&AppEnvironment=datacollector-binary&UploadType=eteventstream
```

There is no way to reroute this domain through Engine.ini, so without client modification you need to use hosts file.

### Helpful official source code
UT4 Source code comes in handy here because we can figure out all possible types of events that get sent out by studying file `UnrealTournament/Source/UnrealTournament/Private/UTAnalytics.cpp`. 

reading `Documents\UnrealTournament\Saved\Stats\stats.html` is also useful as it contains all names and meanings of retrieved stat keys. This file is created after you look at stats inside the game.

### Desired functionality
`statsType` in stats retrieval endpoint is either `alltime`, `monthly`, `weekly`, `daily`. There should probably be buckets for each of these. So all analitycs should at first be saved in a `daily` bucket, then after 24h all these should be moved to `weekly` bucket and so on. then at retrieval of say `montly` we would combine and return stats from `daily`, `weekly` and `monthly`.

### Resources
I am also providing 2 fiddler classic session archives which contain all requests and responses to and from official epicgames servers.
[1) stats and gameplay](https://www.dropbox.com/s/a4oq19p1hyh19sz/ut4-stats-and-gameplay.saz?dl=1)
[1) just gameplay](https://www.dropbox.com/s/a4oq19p1hyh19sz/ut4-stats-and-gameplay.saz?dl=1)

Saibamen:
URLs for resources are this same

[...]

PR #54, opened 2023-01-07
Issue #22: Handling statistics for weapons

belmirp:
The feature works as follows:

- statistics are posted to endpoint: **/ut/api/stats/accountId/{accountId}/bulk?ownertype=1** after each match
- posted statistics are saved with **window=daily** flag
- you can get current account's statistics through **/ut/api/stats/accountId/{accountId}/bulk/window/{window}** endpoint
- daily/weekly/monthly statistics are calculated on-the-fly each time a request is made to GET endpoints
- all-time statistics are calculated each time a new daily statistic is added and saved to the DB, so when requested they are fetched from the DB

image


UT4MasterServer/Services/DatabaseContext.cs
```diff
@@ -13,5 +13,13 @@ public DatabaseContext(IOptions<DatabaseSettings> settings)
...
 		client = new MongoClient(settings.Value.ConnectionString);
 		Database = client.GetDatabase(settings.Value.DatabaseName);
+
+		var statisticsCollection = Database.GetCollection<Statistic>("statistics");
+		var statisticsIndexes = new List<CreateIndexModel<Statistic>>()
+		{
+			new CreateIndexModel<Statistic>(Builders<Statistic>.IndexKeys.Ascending(indexKey => indexKey.AccountId)),
+			new CreateIndexModel<Statistic>(Builders<Statistic>.IndexKeys.Ascending(indexKey => indexKey.CreatedAt))
+		};
+		statisticsCollection.Indexes.CreateMany(statisticsIndexes);
```
timiimit, line 23: I'm not familiar with indexes... is it intentional to create them upon EVERY access to database? new instance of DatabaseContext is created on each request. what does this exactly do? i think this would be better suited to be placed in StatisticsService.

UT4MasterServer/Services/StatisticsService.cs
```diff
@@ -0,0 +1,2203 @@
...
+using UT4MasterServer.Other;
+
+namespace UT4MasterServer.Services;
+
+public sealed class StatisticsService
+{
+	private readonly ILogger<StatisticsService> logger;
+	private readonly IMongoCollection<Statistic> statisticsCollection;
+	private readonly bool allowPasswordGrant;
+
```
timiimit, line 16: `allowPasswordGrant` is never used. it can be removed.

UT4MasterServer/Services/StatisticsService.cs
```diff
@@ -0,0 +1,2203 @@
...
+		DatabaseContext dbContext,
+		IOptions<DatabaseSettings> settings)
+	{
+		this.logger = logger;
+		statisticsCollection = dbContext.Database.GetCollection<Statistic>("statistics");
+		allowPasswordGrant = settings.Value.AllowPasswordGrantType;
+	}
+
+	public async Task<List<StatisticDto>> GetDailyAccountStatistics(string accountId)
+	{
```
timiimit, line 28: just like everywhere else `accountId` should be of type `EpicID`.

im not sure if this method is used, but two of them below look like copy-paste. cant they my merged into one?
`GetAggregateAccountStatistics` already seems like it can be used for any windows of statistics, no?

UT4MasterServer/Services/StatisticsService.cs
```diff
@@ -0,0 +1,2203 @@
...
+		logger.LogInformation("Creating statistics for account: {AccountId}.", accountId);
+
+		var epicAccountId = EpicID.FromString(accountId).ToString();
+
+		var newStatistics = GenerateStatistics(epicAccountId, ownerType, statisticBulkDto);
+
+		await statisticsCollection.InsertManyAsync(newStatistics);
+
+		await UpdateAllTimeAccountStatistics(epicAccountId.ToString(), newStatistics);
```
timiimit, line 139: Remove spaces between calls

UT4MasterServer/Services/StatisticsService.cs
```diff
@@ -0,0 +1,2203 @@
...
+				Value = statisticBulkDto.SupporterScore.Value,
+				Window = StatisticWindow.Daily,
+				OwnerType = ownerType
+			});
+		}
+
+		if (statisticBulkDto.TeamKills.HasValue)
+		{
+			newStatistics.Add(new Statistic()
+			{
```
timiimit, line 2124: This copy paste is alright for now, but this definitely needs some improvement in the future...

UT4MasterServer/Enums/StatisticWindow.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace UT4MasterServer.Enums;
+
+public enum StatisticWindow
```
timiimit, line 3:
```suggestion
public enum StatisticWindow : int
```
if this is being used as an integer representing number of days, it could inherit from int

timiimit (changes requested):
Looks alright overall, but i would perfer some changes before merging. I left comments on the code.

Account ids should always be `EpicID` type.

UT4MasterServer/Controllers/StatisticsController.cs
```diff
@@ -0,0 +1,70 @@
...
+
+namespace UT4MasterServer.Controllers;
+
+[ApiController]
+[Route("ut/api/stats")]
+//[AuthorizeBearer]
+[Produces("application/json")]
+public sealed class StatisticsController : JsonAPIController
+{
+	private readonly ILogger<StatisticsController> logger;
```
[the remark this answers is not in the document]
belmirp: Didn't see it, I will move it there.

UT4MasterServer/DTOs/StatisticBulkDto.cs
```diff
@@ -0,0 +1,207 @@
+﻿namespace UT4MasterServer.DTOs;
+
+public sealed class StatisticBulkDto
+{
```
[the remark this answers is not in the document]
belmirp: DTO stands for Data Transfer Object. It is like ViewModel but more generic. Models should represent only database tables/fields. DTOs should represent requests or responses, or any data that is passed throughout the application, so only needed fields, not the whole model. You should decide whether to use DTOs or ViewModels, in my opinion DTO is more appropriate.

UT4MasterServer/DTOs/StatisticDto.cs
```diff
@@ -0,0 +1,11 @@
+﻿using UT4MasterServer.Enums;
+
+namespace UT4MasterServer.DTOs;
+
+public sealed class StatisticDto
+{
```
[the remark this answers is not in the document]
belmirp: Same response as above.

UT4MasterServer/Models/Statistic.cs
```diff
@@ -0,0 +1,33 @@
...
+namespace UT4MasterServer.Models;
+
+public sealed class Statistic
+{
+	[BsonId]
+	[BsonElement("_id"), BsonRepresentation(BsonType.ObjectId)]
+	public string Id { get; set; } = default!;
+
+	[BsonElement("accountId")]
+	public string AccountId { get; set; } = string.Empty;
```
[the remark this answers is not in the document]
belmirp: I will try to replace string with EpidID type.

UT4MasterServer/Services/StatisticsService.cs
belmirp: The methods were different, but I changed something. I will merge them into one.

UT4MasterServer/Services/StatisticsService.cs
belmirp: It is possible to do it in one foreach loop using reflection, but that is way slower. I don't recommend it.

UT4MasterServer/Enums/StatisticWindow.cs
belmirp: Every enum is of int type by default, no need to mention it.

UT4MasterServer/DTOs/StatisticBulkDto.cs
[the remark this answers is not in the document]
timiimit: alright, lets leave the naming as is

UT4MasterServer/Models/Statistic.cs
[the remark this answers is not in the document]
timiimit: EpicID is automatically converted to strings when serialized to bson. you can take a look at other model classes for example

UT4MasterServer/Services/StatisticsService.cs
timiimit: Yea, definitely don't want to use reflection. I just thought i'd comment about it.

UT4MasterServer/Controllers/UnrealTournamentStatsController.cs
```diff
@@ -15,68 +17,71 @@ namespace UT4MasterServer.Controllers;
...
-	public IActionResult Stats(string id, string leagueName, string category)
+	// Examples:
+	// /ut/api/stats/accountId/0b0f09b400854b9b98932dd9e5abe7c5/bulk/window/daily
+	[HttpGet("accountId/{id}/bulk/window/daily")]
+	public async Task<IActionResult> GetDailyAccountStatistics(string id)
 	{
-		// TODO: Check id and return ErrorResponse if invalid: "Could not convert fakeAccountId to a UUID because it was not in a valid format: Invalid UUID string: fakeAccountId" (errors.com.epicgames.modules.stats.invalid_account_id)
-
-		// TODO: return Stats
+		var accountId = EpicID.FromString(id);
```
timiimit, line 33: Leave comment that you didn't handle
```suggestion
	// TODO: Check id and return ErrorResponse if invalid: "Could not convert fakeAccountId to a UUID because it was not in a valid format: Invalid UUID string: fakeAccountId" (errors.com.epicgames.modules.stats.invalid_account_id)
	var accountId = EpicID.FromString(id);
```

UT4MasterServer/Models/Statistic.cs
```diff
@@ -0,0 +1,33 @@
...
+using MongoDB.Bson.Serialization.Attributes;
+using UT4MasterServer.Enums;
+
+namespace UT4MasterServer.Models;
+
+public sealed class Statistic
+{
+	[BsonId]
+	[BsonElement("_id"), BsonRepresentation(BsonType.ObjectId)]
+	public string Id { get; set; } = default!;
```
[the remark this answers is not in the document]
belmirp: Sure. This is out of habit. There are some static analyzers that warn you to use PascalCase everywhere, so ID would be Id, DTO would be Dto, etc.
I will change it to ID.

UT4MasterServer/Services/DatabaseContext.cs
belmirp: The command is checking if the index with same definition and name already exists, so they only created once, but checked every time.
I will see to put this to another place, and run it only once when application starts.

timiimit (approved):

UT4MasterServer/Services/DatabaseContext.cs
timiimit: I already do something like this in CloudStorageService

timiimit (approved):

PR #59, opened 2023-01-09
Statistics flat structure

belmirp:
This implementation will save the statistics to the DB in flat structure i.e. each statistic will be saved as additional property instead of having one row per statistic.

Also, when querying the statistics for an account, you will get only those that are greater than 0. Empty statistics won't be returned.

This is how each request is saved:
image

And this is how all-time statistics record look in the DB:
image

And this is how GET response looks:
image


Saibamen:
Related: #22

timiimit (approved):

PR #68, opened 2023-01-13
Statistics validation

belmirp:
This is an initial implementation of statistics validation. When a suspicious value is posted as a statistic, that daily record of statistic will be flagged and the list of suspicious properties will be inserted with it for further analysis. Flagged statistics are still included in the daily/weekly/monthly/all-time statistics.

Additional rules and calculations should be added if needed, but they all require further in-game testing.

### DIFF
--- a/UT4MasterServer/Models/Statistic.cs
+++ b/UT4MasterServer/Models/Statistic.cs
@@ -24,4 +24,10 @@ public sealed class Statistic : StatisticBase
 
 	[BsonElement("window")]
 	public StatisticWindow Window { get; set; } = StatisticWindow.Daily;
+
+	[BsonElement("flagged"), BsonIgnoreIfNull]
+	public bool? Flagged { get; set; }
+
+	[BsonElement("flaggedProperties"), BsonIgnoreIfNull]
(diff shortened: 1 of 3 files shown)


UT4MasterServer/Models/Statistic.cs
```diff
@@ -24,4 +24,10 @@ public sealed class Statistic : StatisticBase
 
 	[BsonElement("window")]
 	public StatisticWindow Window { get; set; } = StatisticWindow.Daily;
+
+	[BsonElement("flagged"), BsonIgnoreIfNull]
+	public bool? Flagged { get; set; }
```
timiimit, line 29: Is there a reason for storing boolean?
`FlaggedProperties` can already have null which represents `Flagged` being false.

belmirp: I was actually thinking of that, the existance of **flaggedProperties** can be use to determine if statistic is flagged or not.
I will remove the boolean property and leave only list property, and rename it to **flagged**.

timiimit (approved):

Issue #65, opened 2023-01-12
Implement `InvalidEpicIDException`


timiimit:
EpicID can throw an exception in case an invalid string is passed in when trying to create it. There should be a global exception handler which can catch this error and send out an appropriate response.

belmirp:
@timiimit you can assign this issue to me.

timiimit:
Closed by #71

PR #71, opened 2023-01-13
Global exception handler

belmirp:
This is a basic implementation of global exception handler. In case where exception is not caught, **ErrorsController** will catch it and return only status and message, so that stack trace and other possibly sensitive data is not revealed to the user. We can expand this implementation to return object of different structure or expose more information.

Related: [#65](https://github.com/timiimit/UT4MasterServer/issues/65)


UT4MasterServer/Controllers/ErrorsController.cs
```diff
@@ -0,0 +1,32 @@
+﻿using Microsoft.AspNetCore.Diagnostics;
+using Microsoft.AspNetCore.Mvc;
+using UT4MasterServer.Exceptions;
+
+namespace UT4MasterServer.Controllers;
+
+[ApiController, ApiExplorerSettings(IgnoreApi = true)]
+[Route("api/errors")]
+public class ErrorsController : Controller
```
timiimit, line 9: Does this have to inherit `Controller` or would `ControllerBase` suffice?

belmirp: No, it is not necessary to inherit Controller. I will change it.

timiimit (approved):

PR #76, opened 2023-01-14
Various front end bug fixes

caseyswilliams:
- Tabindex and autofocus in login page for easier entry
- Displaying error messages returned from server responses rather than default messages
- Verifying auth token by querying the verify endpoint rather than checking the expiry in localStorage. Failing this verification clears session from localStorage. Also, made logOut always clear session from localStorage so you can't get stuck logged in. At some point there should probably be a polling of the verify call added to automatically log a user out (or maybe automatically log out when any 401 is received from the API?), but as of now it happens on every page load.

### DIFF
--- a/UT4MasterServer/Controllers/SessionController.cs
+++ b/UT4MasterServer/Controllers/SessionController.cs
@@ -379,7 +379,7 @@ public class SessionController : JsonAPIController
 
 		var obj = new JObject()
 		{
-			{ "token", user.AccessToken },
+			{ "access_token", user.AccessToken },
 			{ "session_id", user.Session.ID.ToString() },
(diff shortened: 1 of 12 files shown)


UT4MasterServer/Controllers/SessionController.cs
```diff
@@ -379,7 +379,7 @@ public async Task<IActionResult> Verify()
 
 		var obj = new JObject()
 		{
-			{ "token", user.AccessToken },
+			{ "access_token", user.AccessToken },
```
timiimit, line 382: having `token` here is actually not a mistake. that's what epic's servers respond with, so it should stay as is.

timiimit (changes requested):

timiimit (approved):

PR #77, opened 2023-01-14
Statistics Controller using ControllerBase and NewtonsoftJson

belmirp:


belmirp:
We decided to use another approach to deal with JSON deserialization error.

PR #79, opened 2023-01-15
Attempting to use an error response instead of throw if ID is not valid

CounterRebel:


timiimit:
The TODO comment you are trying to handle states that it should throw `account_not_found`, the same response as when the account is not found.
I don't think checking for states that are being checked again in constructor is what we want. In case something changes in the future inside `EpicID` those conditions would need to change as well. Better approach would be to just `catch` `InvalidEpicIDException` when it is thrown.

UT4MasterServer/Controllers/AccountController.cs
```diff
@@ -50,6 +50,32 @@ public async Task<IActionResult> GetAccount(string id)
...
+			{
+				ErrorCode = "errors.com.epicgames.account.id_not_hexadecimal",
+				ErrorMessage = "ID needs to be a hexadecimal string",
+				MessageVars = new[] { id },
+				NumericErrorCode = 18008,
+				OriginatingService = "com.epicgames.account.public",
+				Intent = "prod",
+			});
+		}
 		EpicID eid = EpicID.FromString(id);
```
belmirp, line 79: EpicID.FromString will throw and error for the two checks that you are doing above.
If you need additional data, expand InvalidEpicIDException and put it there, and then form the message and this object in ErrorsController.

PR #82, opened 2023-01-15
EpicID error handling

belmirp:
Returning `ErrorResponse `instead of only message when `InvalidEpicIDException `is thrown:

Examples:

`{
  "errorCode": "errors.com.epicgames.account.id_not_32_characters",
  "errorMessage": "ID needs to be 32 characters long.",
  "messageVars": [
    "9c808754479eaa36e7efe896f204da1"
  ],
  "numericErrorCode": 18009,
  "originatingService": null,
  "intent": null,
  "error_description": null,
  "error": null
}`

`{
  "errorCode": "errors.com.epicgames.account.id_not_hexadecimal",
  "errorMessage": "ID needs to be a hexadecimal string.",
  "messageVars": [
    "9c808754479eaa36e7efe896f204da1_"
  ],
  "numericErrorCode": 18008,
  "originatingService": null,
  "intent": null,
  "error_description": null,
  "error": null
}`

### DIFF
--- a/UT4MasterServer/Controllers/ErrorsController.cs
+++ b/UT4MasterServer/Controllers/ErrorsController.cs
@@ -9,6 +9,12 @@ namespace UT4MasterServer.Controllers;
 [Route("api/errors")]
 public class ErrorsController : ControllerBase
 {
+	private readonly ILogger<ErrorsController> logger;
+	public ErrorsController(ILogger<ErrorsController> logger)
+	{
+		this.logger = logger;
+	}
+
 	[HttpGet]
 	public IActionResult Index()
 	{
@@ -16,10 +22,13 @@ public class ErrorsController : ControllerBase
 		var statusCode = 500;
 
 		var exceptionHandlerFeature = HttpContext.Features.Get<IExceptionHandlerPathFeature>();
+		var exception = exceptionHandlerFeature?.Error;
 
-		if (exceptionHandlerFeature?.Error is { } ex)
+		logger.LogError(exception, "Internal server error occurred.");
+
+		if (exception is not null)
 		{
-			switch (ex)
+			switch (exception)
 			{
 				case InvalidEpicIDException invalidEpicIDException:
 					return StatusCode(400, new ErrorResponse()
--- a/UT4MasterServer/Controllers/ErrorsController.cs
+++ b/UT4MasterServer/Controllers/ErrorsController.cs
@@ -9,7 +9,10 @@ namespace UT4MasterServer.Controllers;
 [Route("api/errors")]
 public class ErrorsController : ControllerBase
 {
+	private const string InternalServerError = "Internal server error occurred.";
+
 	private readonly ILogger<ErrorsController> logger;
+
 	public ErrorsController(ILogger<ErrorsController> logger)
 	{
 		this.logger = logger;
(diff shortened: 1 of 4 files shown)


UT4MasterServer/Other/EpicID.cs
```diff
@@ -27,14 +27,14 @@ private EpicID(string id)
 	{
 		if (id.Length != 32)
 		{
-			throw new InvalidEpicIDException("ID needs to be 32 characters long.");
+			throw new InvalidEpicIDException("ID needs to be 32 characters long.", id, 18009, "errors.com.epicgames.account.id_not_32_Characters");
```
Saibamen, line 30:
```suggestion
			throw new InvalidEpicIDException("ID needs to be 32 characters long.", id, 18009, "errors.com.epicgames.account.id_not_32_characters");
```

belmirp: Copy-paste error, fixed it.

UT4MasterServer/Controllers/ErrorsController.cs
```diff
@@ -20,10 +21,14 @@ public IActionResult Index()
 		{
```
Saibamen, line 21: Can you also add logging exception here? `logger.LogError(ex)`

belmirp: Added, here is an example:
image

UT4MasterServer/Controllers/ErrorsController.cs
```diff
@@ -1,29 +1,43 @@
...
 	[HttpGet]
 	public IActionResult Index()
 	{
 		var message = "Internal server error occurred.";
 		var statusCode = 500;
 
 		var exceptionHandlerFeature = HttpContext.Features.Get<IExceptionHandlerPathFeature>();
+		var exception = exceptionHandlerFeature?.Error;
+
+		logger.LogError(exception, "Internal server error occurred.");
```
Saibamen, line 27: We already have this string in `message` variable.
```suggestion
		logger.LogError(exception, message);
```

belmirp: I know, but VS is complaining about that:
image

I will store the text in a constant.

UT4MasterServer/Exceptions/InvalidEpicIDException.cs
```diff
@@ -5,7 +5,47 @@ namespace UT4MasterServer.Exceptions;
...
-	public InvalidEpicIDException(string message) : base(message) { }
+	private string _id { get; set; }
+	public string ID { get { return _id; } }
 
-	protected InvalidEpicIDException(SerializationInfo serializationInfo, StreamingContext streamingContext) : base(serializationInfo, streamingContext) { }
+	private int _numericErrorCode { get; set; }
+	public int NumericErrorCode { get { return _numericErrorCode; } }
+
+	private string _errorCode { get; set; }
+	public string ErrorCode { get { return _errorCode; } }
```
timiimit, line 15: Is there a reason for having a set of private/public properties? public property could just have setter marked as private

UT4MasterServer/Other/EpicID.cs
```diff
@@ -27,14 +27,14 @@ private EpicID(string id)
 	{
 		if (id.Length != 32)
 		{
-			throw new InvalidEpicIDException("ID needs to be 32 characters long.");
+			throw new InvalidEpicIDException("ID needs to be 32 characters long.", id, 18009, "errors.com.epicgames.account.id_not_32_characters");
```
timiimit, line 30: are the numerical error codes taken from somewhere or did you come up with them?
i think it should be the same as original when `errors.com.epicgames.modules.stats.invalid_account_id` or similar error is triggered on official servers

timiimit (approved):

PR #87, opened 2023-01-16
Stats front end

caseyswilliams:
This is a preliminary take at the stats front end. I didn't have any test data to validate that it works as expected, the front end models may need to be tweaked to match what is returned from the API. 
There is also no weapon death pie charts as exist in the game UI, just because I didn't want to overcomplicate the first pass.
The user selection is just a standard HTML select component at the moment, but should be replaced with something more user friendly at some point. It should be usable enough while the number of registered players is small.


UT4MasterServer/Controllers/AccountController.cs
```diff
@@ -141,6 +141,29 @@ public async Task<IActionResult> GetAccounts([FromQuery(Name = "accountId")] Lis
 		return Json(arr);
 	}
 
+	[HttpGet("public/accounts")]
+	public async Task<IActionResult> GetAllAccounts()
+	{
```
timiimit, line 146: Its fine having this endpoint here for now, but perhaps it would be a better fit in PersonaController

timiimit (approved):

PR #89, opened 2023-01-16
noid: during docker build command throws an error like this CSC : err…

nrodear:
…or CS5001: Program does not contain a static 'Main' method suitable for an entry point

steps to reproduce
on linux only
cd ./UT4MasterServer/UT4MasterServer
docker build .
...
then error CS5001: Program does not contain 
after apply patch 
docker build .
..
build ok


Saibamen:
Everything is OK. Our GH Actions can build Docker image for UT4MasterServer

timiimit:
I would actually like to merge this so that Dockerfile is located at the same relative location as in `UT4MasterServer.Web` but i don't think visual studio (the main IDE currently used for development) can handle this. I will test it out as soon as i have time and merge if it does in fact work.

If you just want to run it on linux, the way you can run the whole project is by using `docker-compose up` inside the root directory. I have no problems running it this way on live website, but do let me know if it does not work for you.

timiimit:
I triggered GH action which checks if everything compiles and it failed. So there definitely something isnt right with your changes.

.github/workflows/docker_build.yml
```diff
@@ -25,6 +25,6 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: Build Docker Image
-        run: docker build --pull -t ${{ env.IMAGE_NAME }} -f ./UT4MasterServer/Dockerfile .
+        run: docker build --pull -t ${{ env.IMAGE_NAME }} -f ./UT4MasterServer/.docker/Dockerfile .
```
nrodear, line 28: this file move. now the git build should work.

nrodear:
> I triggered GH action which checks if everything compiles and it failed. So there definitely something isnt right with your changes.

i fixed it in my last commit below. works now.

timiimit:
I saw you changed it. I'm sorry its taking a while, i'm really busy this whole week.
I still havent had the time to test it in visual studio. Will probably do that tomorrow.
I'll merge it then if it works or let you know if it doesn't.

nrodear:
why this merge request? maybe you dont needed and feel free to delete merge request.
why this merge request? 
so what happen for me to test the build and docker? 
- i clone the code
- go into subfolder and build without any ide or ui , only console
 
 docker build .
 
 this didnt work and i fixed it. you dont needed, np. :)
 
its no problem for me, i close it.

Saibamen:
You need to use docker-compose

`docker-compose build --pull` and `docker-compose up`

PR #91, opened 2023-01-18
Server List and Player Card

caseyswilliams:
A lot of files but it is mostly images for the avatars and flags.
The Server listing is kind of just a best guess, not sure on the properties, or if my assumptions about what is a server vs a hub vs a match are correct. I will be able to tell better with real data.
The player card is kind of small because I didn't find rank data in the api, that can be added when we have it / when someone tells me where it is.

Note: I'm not in love with the styling of either of these pages, I just used some basic bootstrap stuff to throw it together. I will put more time into making it prettier later, and I'm open to suggestions.


UT4MasterServer/Controllers/UnrealTournamentMatchmakingController.cs
```diff
@@ -255,18 +255,20 @@ public async Task<IActionResult> ListGameServers([FromBody] GameServerFilter fil
...
-		//	new GameServer("cracked", "[DS]dallastn-22938", "192.223.24.243"),
-		//	new GameServer("the", "[DS]dallastn-22938", "192.223.24.243"),
-		//	new GameServer("code", "[DS]dallastn-22938", "192.223.24.243"),
-		//	new GameServer("and", "[DS]dallastn-22938", "192.223.24.243"),
-		//	new GameServer("entered", "[DS]dallastn-22938", "192.223.24.243"),
-		//	new GameServer("the", "[DS]dallastn-22938", "192.223.24.243"), // does not show, due to duplicate data
-		//	new GameServer("matrix", "[DS]dallastn-22938", "192.223.24.243"),
-		//};
-
+#if DEBUG
```
timiimit, line 258: If you leave this uncommented, then please add `//#define SHOW_TEST_HUBS` (notice comment) at the top.
Then change this line (258) to `#if DEBUG && SHOW_TEST_HUBS`

Then uncomment the `#define` when you want to test locally.

UT4MasterServer.Web/src/pages/Profile/PlayerCard.vue
```diff
@@ -1,8 +1,81 @@
 <template>
-  <h1>Player</h1>
-  <h3>Coming Soon!</h3>
-  <!-- 
-    Placeholder for the playerCard UI 
-  https://www.epicgames.com/unrealtournament/en-US/playerCard?playerId=0b0f09b400854b9b98932dd9e5abe7c5
```
timiimit, line 6: Link in this comment could be left in, so that we don't forget about original link after shutdown.

timiimit (changes requested):

timiimit:
you can find some live examples of server data here:
https://master-ut4.timiimit.com/ut/api/matchmaking/session/matchMakingRequest

timiimit:
Also, idk how you got images, but i hope you got them all.

UT4MasterServer.Web/src/types/game-server.ts
```diff
@@ -0,0 +1,57 @@
+export interface IGameServerAttributes {
```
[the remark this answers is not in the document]
caseyswilliams: Since we are referencing these attributes to make decisions (like whether or not it is a hub, etc) I'd like to keep the type safety. I can remove any of the attributes that we aren't using, especially knowing now they may not be constant. Defining an interface in Typescript isn't like doing the same in C#, the JSON doesn't get serialized into the type. Typescript only exists at compile time, it doesn't actually try and create an object conforming to your interface, it is only for code completion and compilation.

UT4MasterServer/Controllers/UnrealTournamentMatchmakingController.cs
caseyswilliams: I'm just gonna revert it, the test data doesn't really work anyway. For testing I just returned the response from the prod server manually. Should probably come up with a better means of getting test data at some point.

UT4MasterServer.Web/src/pages/Profile/PlayerCard.vue
caseyswilliams: Agreed

UT4MasterServer.Web/src/stores/server-store.ts
```diff
@@ -0,0 +1,45 @@
...
+    },
+    get status() {
+        return _status.value;
+    },
+    async fetchGameServers() {
+        try {
+            _status.value = AsyncStatus.BUSY;
+            _allServers.value = await _serverService.getGameServers();
+            const hubMap: Record<string, IGameHub> = {};
+            // TODO: I'm not sure about this logic for determining Hubs, there may be a better way to handle this once I see real data
```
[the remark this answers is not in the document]
caseyswilliams: Updated

UT4MasterServer.Web/src/components/HubListing.vue
```diff
@@ -0,0 +1,36 @@
+<template>
+    <!-- TODO: Not really sure about any of these properties, hard to tell from test data -->
+    <div class="list-group">
+        <a class="list-group-item list-group-item-action active">
+            Matches: {{ hub.serverName }}
```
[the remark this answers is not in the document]
caseyswilliams: Updated

caseyswilliams:
> Also, idk how you got images, but i hope you got them all.

I _think_ I got them all. I got them from the epic static site, where they were referenced from in the playerCard. For the flags I went through the dropdown and saved each option. The avatar I just tried replacing the numbers at the end until I got to one that 404ed.

timiimit (approved):

PR #92, opened 2023-01-18
Add Serilog

### LINKED ISSUE #83: Add logging to text file
Currently there is only console logging. There should be some way to view logs externally. The simplest solution is a log file. This file should be stored outside of container for persistance, same way as db.

Saibamen:
Fixes #83 

Filename example: `log-20230118.log`

Log template: `"{Timestamp:yyyy-MM-dd HH:mm:ss zzz} [{Level}] ({ThreadId}) {RequestId}-{SourceContext}: {Message}{NewLine}{Exception}"`

Log example:
`
2023-01-18 09:21:13 +00:00 [Information] (20) -UT4MasterServer.Services.ApplicationStartupService: Background task deleted 0 expired sessions.
`

### DIFF
--- a/UT4MasterServer/Program.cs
+++ b/UT4MasterServer/Program.cs
@@ -1,8 +1,10 @@
 using Microsoft.AspNetCore.Authentication;
 using Microsoft.OpenApi.Models;
 using MongoDB.Bson.Serialization;
+using Serilog;
 using System.Net;
 using UT4MasterServer.Authentication;
+using UT4MasterServer.Configuration;
 using UT4MasterServer.Models;
 using UT4MasterServer.Other;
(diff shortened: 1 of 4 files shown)


UT4MasterServer/Program.cs
```diff
@@ -153,6 +153,11 @@ public static void Main(string[] args)
 
 		var app = builder.Build();
 
+		IConfiguration configuration = app.Configuration;
+		IWebHostEnvironment environment = app.Environment;
```
timiimit, line 157: There is no need to have these two variables since there's only one use of them.

Saibamen: Done

timiimit (approved):

PR #93, opened 2023-01-19
Minor UI tweaks and username bug fix

caseyswilliams:
- Fixed bug where username value sticks between logins of different users
- Added a CopyButton component that has some visual feedback for the user that the content was copied
- Added empty states on Servers page for the case with no hubs or selected hub has no matches


UT4MasterServer.Web/src/services/authentication.service.ts
```diff
@@ -81,6 +82,7 @@ export default class AuthenticationService extends HttpService {
             console.error('Error killing session', err);
         }
         finally {
+            AccountStore.account = null;
```
caseyswilliams, line 85: This is the bug fix, needed to clear the account on log out, otherwise it doesn't fetch it again and retains the account of the first user to log in

Saibamen: Can you write a helper function with `AccountStore.Clear()` and `SessionStore.Clear()` to have it in one place and able to use it in for example when returning `UNAUTHORIZED` error from backend?

caseyswilliams: That would be this function. This is the one place.

caseyswilliams: I thought about having a clear function on the stores, but each only has the one prop to clear, so I left it out for now. As for the single place to call in order to end the session and log the user out, that would be this function. Having clear functions on the stores wouldn't be one place to accomplish this, it would be two.

caseyswilliams: As for logging out when receiving a 401 from an API request, I'm not sure that is always the thing to do. There may be some calls that it would make sense for, but I could see cases where a 401 response would just be the normal way to tell a user they weren't allowed to do something and it shouldn't log them out then. I decided to just not do any automatic logging out based on 401s for now, that could change at some point.

UT4MasterServer.Web/src/pages/Servers.vue
```diff
@@ -5,8 +5,9 @@
         <div class="list-group">
           <a class="list-group-item list-group-item-action active">Hubs</a>
           <Hub :hub="hub" v-for="hub in ServerStore.hubs" :selected-hub="selectedHub" @click="viewHub(hub)" />
+          <Hub v-if="!ServerStore.hubs.length" />
         </div>
-        <!-- TODO: Should also have Servers listed here in addition to Hubs, but I'm not yet sure how to differentiate -->
+        <!-- TODO: Should(?) also have Servers listed here in addition to Hubs, but I'm not yet sure how to differentiate -->
```
timiimit, line 10: There should be a way to switch between listing hubs and servers eventually. same as in the game. (maybe additionally even qp matches once we get there)

timiimit (approved):

PR #100, opened 2023-01-21
Delete old statistic records

belmirp:
By default, this method is deleting statistics that are at least 30 days old and they don't have flagged properties. Currently, it can be only called from controller. When decided, we should put call to this method on specific places such as: after posting new statistics, after querying statistics, etc.

### DIFF
--- a/UT4MasterServer/Controllers/UnrealTournamentStatsController.cs
+++ b/UT4MasterServer/Controllers/UnrealTournamentStatsController.cs
@@ -96,11 +96,4 @@ public class UnrealTournamentStatsController : JsonAPIController
 		await statisticsService.CreateAccountStatisticsAsync(accountId, ownerType, statisticBase);
 		return Ok();
 	}
-
-	[HttpDelete]
-	public async Task<IActionResult> DeleteOldStatistics(int days = 30, bool skipFlagged = true)
-	{
-		var result = await statisticsService.DeleteOldStatisticsAsync(days, skipFlagged);
-		return Ok(result);
-	}
 }
--- a/UT4MasterServer/Controllers/UnrealTournamentStatsController.cs
+++ b/UT4MasterServer/Controllers/UnrealTournamentStatsController.cs
@@ -96,4 +96,11 @@ public class UnrealTournamentStatsController : JsonAPIController
 		await statisticsService.CreateAccountStatisticsAsync(accountId, ownerType, statisticBase);
 		return Ok();
 	}
+
+	[HttpDelete]
+	public async Task<IActionResult> DeleteOldStatistics(int days = 30, bool skipFlagged = true)
+	{
+		var result = await statisticsService.DeleteOldStatisticsAsync(days, skipFlagged);
+		return Ok(result);
+	}
 }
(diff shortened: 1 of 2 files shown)


UT4MasterServer/Services/StatisticsService.cs
```diff
@@ -490,5 +492,36 @@ private static Statistic MergeStatistics(List<Statistic> statistics)
...
+			throw new ArgumentException($"You can delete only statistics that are at least {MinimumDaysForDeletingOldStatistics} days old.", nameof(days));
+		}
+
+		var removeBeforeDate = DateTime.UtcNow.Date.AddDays(-days);
+
+		var filter = Builders<Statistic>.Filter.Lt(f => f.CreatedAt, removeBeforeDate);
+		
+		if (skipFlagged)
+		{
+			filter &= Builders<Statistic>.Filter.Eq(f => f.Flagged, null);
```
timiimit, line 516: This works? Either way, I would prefer the use of `Filter.Exists`.

UT4MasterServer/Controllers/UnrealTournamentStatsController.cs
```diff
@@ -96,4 +96,11 @@ public async Task<IActionResult> GetAllTimeAccountStatistics(string id)
 		await statisticsService.CreateAccountStatisticsAsync(accountId, ownerType, statisticBase);
 		return Ok();
 	}
+
+	[HttpDelete]
+	public async Task<IActionResult> DeleteOldStatistics(int days = 30, bool skipFlagged = true)
+	{
+		var result = await statisticsService.DeleteOldStatisticsAsync(days, skipFlagged);
+		return Ok(result);
+	}
```
timiimit, line 105: There have to be some extra authorization checks somewhere. Please wait until #102 merges and then make sure account has sufficient permissions to run this action.

timiimit (changes requested):

UT4MasterServer/Services/StatisticsService.cs
belmirp: It works, but I will replace it with Exists.

UT4MasterServer/Controllers/UnrealTournamentStatsController.cs
belmirp: Agreed to remove this endpoint until needed.

UT4MasterServer/Services/StatisticsService.cs
```diff
@@ -509,11 +509,12 @@ public async Task<long> DeleteOldStatisticsAsync(int days = 30, bool skipFlagged
 
 		var removeBeforeDate = DateTime.UtcNow.Date.AddDays(-days);
 
-		var filter = Builders<Statistic>.Filter.Lt(f => f.CreatedAt, removeBeforeDate);
-		
+		var filter = Builders<Statistic>.Filter.Lt(f => f.CreatedAt, removeBeforeDate) &
+					 Builders<Statistic>.Filter.In(f => f.Window, new List<StatisticWindow>() { StatisticWindow.Daily });
```
timiimit, line 513: use `Eq` instead of `In` if just testing if its one specific value

timiimit (changes requested):

UT4MasterServer/Services/StatisticsService.cs
belmirp: That is left because I will remove both Daily and DailyMerged statistics, but this PR doesn't have DailyMerged enum member yet.

timiimit (approved):

PR #102, opened 2023-01-21
Account and GameServer permissions

timiimit:
- Add simple permissions to account
- Add `clients` collection to db which lists all allowed client ID:Secret authorization pairs
- Add `trustedservers` collection listing all client ids which are to be considered trusted server instances
- Try to mimic epic's logic when handling GameServers instead of a secure logic

### DIFF
--- a/UT4MasterServer/Authentication/BasicAuthenticationHandler.cs
+++ b/UT4MasterServer/Authentication/BasicAuthenticationHandler.cs
@@ -32,12 +32,12 @@ public class BasicAuthenticationHandler : AuthenticationHandler<AuthenticationSc
 			}
 
 			var clientParser = new ClientIdentification(authorization.Value);
-			if (clientParser.ID.IsEmpty || clientParser.Secret.IsEmpty)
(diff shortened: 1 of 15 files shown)


UT4MasterServer/Authentication/BasicAuthenticationHandler.cs
```diff
@@ -26,17 +31,25 @@ protected override async Task<AuthenticateResult> HandleAuthenticateAsync()
 				return AuthenticateResult.Fail("unexpected scheme");
 			}
 
-			var client = new ClientIdentification(authorization.Value);
-			if (client.ID.IsEmpty || client.Secret.IsEmpty)
+			var clientParser = new ClientIdentification(authorization.Value);
+			if (clientParser.ID.IsEmpty || string.IsNullOrEmpty(clientParser.Secret))
```
Saibamen, line 35: IMO better is `IsNullOrWhitespace`

timiimit: I was also considering this, but i ultimately decided on `IsNullOrEmpty` because if we create a Client with just whitespace (very unlikely) as secret we still want to handle it properly.

PR #107, opened 2023-01-22
Merge old statistic records into one

belmirp:
This job will be executed once per day at the time defined in the appsettings.

It will look for daily non-flagged statistic records that are older than 7 days, and merge them into one statistic record with `Window = DailyMerged`. If there is only one record per account per day, that record is only modified to have `Window = DailyMerged`.

Here are some examples of the process:

MergingOldStatistics_Example1
MergingOldStatistics_Example2
MergingOldStatistics_Example3

### DIFF
--- a/UT4MasterServer/Models/StatisticBase.cs
+++ b/UT4MasterServer/Models/StatisticBase.cs
@@ -919,25 +919,19 @@ public class StatisticBase
 
 	private void ValidateMultiAndSpreeKillStatistic(int? value, int multiplier, string statisticName, List<string> flaggedFields)
 	{
-		if (value.HasValue)
+		if (value.HasValue && (!Kills.HasValue || (Kills.HasValue && value.Value * multiplier > Kills.Value)))
 		{
-			if (!Kills.HasValue || (Kills.HasValue && value.Value * multiplier > Kills.Value))
-			{
-				flaggedFields.Add(statisticName);
-				flaggedFields.Add(nameof(Kills));
-			}
+			flaggedFields.Add(statisticName);
+			flaggedFields.Add(nameof(Kills));
 		}
 	}
 
 	private void ValidateTimeStatistic(double? value, string statisticName, List<string> flaggedFields)
 	{
-		if (value.HasValue)
+		if (value.HasValue && TimePlayed.HasValue && value.Value > TimePlayed.Value)
 		{
-			if (TimePlayed.HasValue && value.Value > TimePlayed.Value)
-			{
-				flaggedFields.Add(statisticName);
-				flaggedFields.Add(nameof(TimePlayed));
-			}
+			flaggedFields.Add(statisticName);
+			flaggedFields.Add(nameof(TimePlayed));
 		}
 	}
 
--- a/UT4MasterServer/Models/StatisticBase.cs
+++ b/UT4MasterServer/Models/StatisticBase.cs
@@ -342,7 +342,7 @@ public class StatisticBase
 			LightningRifleSecondaryKills,
 			RedeemerKills,
 			InstagibKills,
-			TelefragKills
+			TelefragKills,
 		}.Sum() ?? 0;
 
 		#region Quick Look Validations
--- a/UT4MasterServer/Models/StatisticBase.cs
+++ b/UT4MasterServer/Models/StatisticBase.cs
@@ -316,337 +316,6 @@ public class StatisticBase
 		TeamKills = statisticBase.TeamKills;
 	}
 
-	public List<string> Validate()
-	{
-		var flaggedFields = new List<string>();
-
-		var totalKills = new int?[]
-		{
-			ImpactHammerKills,
-			EnforcerKills,
-			BioRifleKills,
-			BioLauncherKills,
-			ShockBeamKills,
-			ShockCoreKills,
-			ShockComboKills,
-			LinkKills,
-			LinkBeamKills,
-			MinigunKills,
-			MinigunShardKills,
-			FlakShardKills,
-			FlakShellKills,
-			RocketKills,
-			SniperKills,
-			SniperHeadshotKills,
-			LightningRiflePrimaryKills,
-			LightningRifleSecondaryKills,
-			RedeemerKills,
-			InstagibKills,
-			TelefragKills,
-		}.Sum() ?? 0;
-
-		#region Quick Look Validations
-
-		// Since this is added after match, this property should be only 1
-		if (MatchesPlayed.HasValue && MatchesPlayed.Value > 1)
-		{
-			flaggedFields.Add(nameof(MatchesPlayed));
-		}
-		// Since this is added after match, this property should be only 1
-		if (MatchesQuit.HasValue && MatchesQuit.Value > 1)
-		{
-			flaggedFields.Add(nameof(MatchesQuit));
-		}
-		// You shouldn't have both Wins and Losses value available at the same time
-		if (MatchesPlayed.HasValue && MatchesQuit.HasValue && MatchesPlayed.Value > 0 && MatchesQuit.Value > 0)
-		{
-			flaggedFields.Add(nameof(MatchesPlayed));
-			flaggedFields.Add(nameof(MatchesQuit));
-		}
-		// Since this is added after match, this property should be only 1
-		if (Wins.HasValue && Wins.Value > 1)
-		{
-			flaggedFields.Add(nameof(Wins));
-		}
-		// Since this is added after match, this property should be only 1
-		if (Losses.HasValue && Losses.Value > 1)
-		{
-			flaggedFields.Add(nameof(Losses));
-		}
-		// You shouldn't have both Wins and Losses value available at the same time
-		if (Wins.HasValue && Losses.HasValue && Wins.Value > 0 && Losses.Value > 0)
-		{
-			flaggedFields.Add(nameof(Wins));
-			flaggedFields.Add(nameof(Losses));
-		}
-		// Maximum time for custom game is 60 minutes (1 minute is added just to be sure)
-		if (TimePlayed.HasValue && TimePlayed.Value > (60 * 60) + 60)
-		{
-			flaggedFields.Add(nameof(TimePlayed));
-		}
-		if (Kills.HasValue)
-		{
-			// Kills are reported if you manage to achieve 1 kill each 1.5 second
-			if (TimePlayed.HasValue && Kills.Value > (TimePlayed.Value / 1.5))
-			{
-				flaggedFields.Add(nameof(Kills));
-				flaggedFields.Add(nameof(TimePlayed));
-			}
-			// Kills are reported if they don't match the sum of individual weapon kills
-			if (Kills.Value != totalKills)
-			{
-				flaggedFields.Add(nameof(Kills));
-			}
-		}
-		// Deaths are reported if you manage to achieve 1 death each 2 seconds
-		if (Deaths.HasValue && TimePlayed.HasValue && Deaths.Value > (TimePlayed.Value / 2))
-		{
-			flaggedFields.Add(nameof(Deaths));
-			flaggedFields.Add(nameof(TimePlayed));
-		}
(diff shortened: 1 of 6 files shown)


UT4MasterServer/Models/StatisticBase.cs
```diff
@@ -516,109 +516,130 @@ public List<string> Validate()
...
 		{
 			flaggedFields.Add(nameof(MinigunKills));
+			flaggedFields.Add(nameof(Kills));
 		}
 		// Minigun Shard kills are reported if they are more than the actual kills
-		if (MinigunShardKills.HasValue && Kills.HasValue && MinigunShardKills.Value > Kills.Value)
+		if (MinigunShardKills.HasValue && (!Kills.HasValue || MinigunShardKills.Value > Kills.Value))
 		{
 			flaggedFields.Add(nameof(MinigunShardKills));
+			flaggedFields.Add(nameof(Kills));
```
Saibamen, line 582: Maybe add a helper method, for example `AddKillsFor` to reduce code duplication

belmirp: This part is adding only name of the property to the list of flagged properties. I am not sure for what exactly would you add extension method?

Saibamen: To reduce code. I updated my comment with code example

belmirp: You are not going to reduce anything. You will replace this:

image

with this:
image

belmirp: If the line count is important, I could do it like this:
image

UT4MasterServer/Services/StatisticsService.cs
```diff
@@ -249,6 +249,125 @@ public async Task CreateAccountStatisticsAsync(EpicID accountID, OwnerType _, St
...
+
+		foreach (var statisticsGroup in statisticsGrouped)
+		{
+			// Statistics that have multiple records per day should be merged together or merged with already merged statistics
+			if (statisticsGroup.Count > 1)
+			{
+				var statisticsToMerge = new List<Statistic>(statisticsGroup.Statistics);
+
+				if (alreadyMergedStatistics.FirstOrDefault(f => f.AccountID == statisticsGroup.AccountId &&
+																f.CreatedAt.Date == statisticsGroup.CreatedAt.Date) is { } alreadyMergedStatistic)
```
Saibamen, line 302: Fix indentation

UT4MasterServer/Models/StatisticBase.cs
Saibamen: But IMHO using `AddRange` will be better for code optimization.

image

(Remember to add comma at the end for better git diff view and PR code review experience in the future

belmirp: The problem with AddRange is that you have to instantiate unnecessary array/list each time. Using Add in multiple lines doesn't allocate anything additional.

UT4MasterServer/Services/ApplicationBackgroundService.cs
```diff
@@ -1,66 +1,87 @@
...
 
-		public void Dispose()
+	/// <summary>
+	/// This job is executed each day, and it merges statistics older than 7 days into single record per day per account
+	/// </summary>
+	private async Task MergeOldStatisticsAsync(IServiceScope scope)
+	{
+		var currentTime = TimeZoneInfo.ConvertTime(DateTimeOffset.UtcNow, TimeZoneInfo.FindSystemTimeZoneById(settings.MergeOldStatisticsTimeZone));
+		if (currentTime.Hour == settings.MergeOldStatisticsHour &&
+		   (!lastDateMergeOldStatisticsExecuted.HasValue || lastDateMergeOldStatisticsExecuted.Value.Date != currentTime.Date))
```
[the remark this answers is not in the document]
belmirp: I actually added that to be in parallel with the condition above. What do you suggest, to stick with the tabs, and have something like this:
image

or could be worse like this:
image

UT4MasterServer/Models/StatisticBase.cs
timiimit: How about just having something like this (provided just as pseudocode):
```cs
	public void ValidateProperty(string name, int? value)
	{
		if (value.HasValue && Kills.HasValue && value.Value > Kills.Value)
		{
			flaggedFields.Add(name);
			flaggedFields.Add(nameof(Kills));
		}
	}
```

timiimit: This would reduce number of lines significantly imo

timiimit: Then create an array with all name/value pairs and just loop though all to reduce perhaps even more

UT4MasterServer/Services/ApplicationBackgroundService.cs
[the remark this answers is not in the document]
timiimit: You can create additional variable separating conditions apart

UT4MasterServer/Models/StatisticBase.cs
belmirp: Not all rules are the same. I wouldn't make multiple methods out of one just for the sake of lines.

timiimit: All weapon stats seem the same. Having ~5 extra methods with different conditions is definitely something i'd prefer over this long unreadable mess. Its not for the sake of lines, its for the sake of maintainability and ease of reading.

UT4MasterServer/Services/StatisticsService.cs
```diff
@@ -249,6 +249,125 @@ public async Task CreateAccountStatisticsAsync(EpicID accountID, OwnerType _, St
...
```
timiimit, line 302: Indentation

UT4MasterServer/Services/StatisticsService.cs
```diff
@@ -249,6 +249,125 @@ public async Task CreateAccountStatisticsAsync(EpicID accountID, OwnerType _, St
...
+
+				statisticIdsToDelete.AddRange(mergedStatisticsIds);
+
+				logger.LogInformation("Merged the following statistics into one record: {StatisticIds}.", string.Join(", ", mergedStatisticsIds));
+			}
+			// Statistics that have one record per day should be either modified or merged with already merged statistics
+			else
+			{
+				if (alreadyMergedStatistics.FirstOrDefault(f => f.AccountID == statisticsGroup.AccountId &&
+																f.CreatedAt.Date == statisticsGroup.CreatedAt.Date) is { } alreadyMergedStatistic)
```
timiimit, line 322: Here too

UT4MasterServer/Models/StatisticBase.cs
belmirp: If that is what you prefer, I will extract similar logic to separate methods.

Saibamen: I run a simple BenchmarkDotNet benchmark and `DoubleAdd` is better

```
|    Method |       Mean |     Error |     StdDev |     Median |        Min |         Max | Ratio | RatioSD | Allocated | Alloc Ratio |
|---------- |-----------:|----------:|-----------:|-----------:|-----------:|------------:|------:|--------:|----------:|------------:|
| DoubleAdd |   510.7 ns |  68.49 ns |   929.4 ns |   400.0 ns |   300.0 ns | 24,700.0 ns |  1.00 |    0.00 |     640 B |        1.00 |
|  AddRange | 2,627.3 ns | 187.31 ns | 2,541.9 ns | 2,100.0 ns | 1,500.0 ns | 25,600.0 ns |  6.21 |    5.98 |     680 B |        1.06 |
```

Benchmark code:

```C#
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Engines;
using BenchmarkDotNet.Running;

namespace Benchmark;

[SimpleJob(RunStrategy.Monitoring, launchCount: 2, warmupCount: 5, iterationCount: 1000)]
[MemoryDiagnoser]
[MinColumn, MaxColumn, MedianColumn]
public class AddTwiceVsAddRange
{
	public int? Wins { get; set; }
	public int? Losses { get; set; }

	private List<string> flaggedFields;

	[GlobalSetup]
	public void GlobalSetup()
	{
		flaggedFields = new List<string>();
	}

	[Benchmark(Baseline = true)]
	public List<string> DoubleAdd()
	{
		flaggedFields.Add(nameof(Wins));
		flaggedFields.Add(nameof(Losses));
		return flaggedFields;
	}

	[Benchmark]
	public List<string> AddRange()
	{
		flaggedFields.AddRange(new[]
		{
			nameof(Wins),
			nameof(Losses),
		});

		return flaggedFields;
	}
}

public class Program
{
	public static void Main(string[] args)
	{
		var summary = BenchmarkRunner.Run<AddTwiceVsAddRange>();
	}
}
```

belmirp: Extracted logic to multiple methods.

UT4MasterServer/Services/StatisticsService.cs
belmirp: Indentation is OK, just not displayed by GitHub correctly. You can check it by viewing the whole file:
image

UT4MasterServer/Services/StatisticsService.cs
belmirp: Same as above.

UT4MasterServer/Models/StatisticBase.cs
```diff
@@ -1110,4 +779,176 @@ public List<string> Validate()
...
+			flaggedFields.Add(statisticName);
+			flaggedFields.Add(nameof(TimePlayed));
+		}
+	}
+
+	private void ValidateMultiAndSpreeKillStatistic(int? value, int multiplier, string statisticName, List<string> flaggedFields)
+	{
+		if (value.HasValue)
+		{
+			if (!Kills.HasValue || (Kills.HasValue && value.Value * multiplier > Kills.Value))
```
Saibamen, line 924: Merge it into one IF statement or return like this to reduce nesting:

```C#
if (!value.HasValue)
{
	return;
}
```

UT4MasterServer/Models/StatisticBase.cs
```diff
@@ -1110,4 +779,176 @@ public List<string> Validate()
...
+				flaggedFields.Add(statisticName);
+				flaggedFields.Add(nameof(Kills));
+			}
+		}
+	}
+
+	private void ValidateTimeStatistic(double? value, string statisticName, List<string> flaggedFields)
+	{
+		if (value.HasValue)
+		{
```
Saibamen, line 935: Same here

Saibamen (changes requested):

UT4MasterServer/Models/StatisticBase.cs
```diff
@@ -1089,4 +779,176 @@ public List<string> Validate()
...
+		ValidateIfBothStatisticsExist(MatchesPlayed, MatchesQuit, nameof(MatchesPlayed), nameof(MatchesQuit), flaggedFields);
+		ValidateIfBothStatisticsExist(Wins, Losses, nameof(Wins), nameof(Losses), flaggedFields);
+
+		// Maximum time for custom game is 60 minutes (1 minute is added just to be sure)
+		if (TimePlayed.HasValue && TimePlayed.Value > (60 * 60) + 60)
+		{
+			flaggedFields.Add(nameof(TimePlayed));
+		}
+		if (Kills.HasValue)
+		{
```
Saibamen, line 830:
```suggestion
		}

		if (Kills.HasValue)
		{
```

UT4MasterServer/Models/StatisticBase.cs
belmirp: Done.

UT4MasterServer/Models/StatisticBase.cs
belmirp: Done.

timiimit (approved):

belmirp:

Issue #33, opened 2022-12-30
Use either System.Text.Json or Newtonsoft.Json across whole project


timiimit:
Currently there are places where using one just seemed simpler than the other or i just couldn't figure one out, so i just jumped between the use of both.

We should decide which one is better for our use cases and go with one throughout whole code.

timiimit:
Get rid of `Newtonsoft.Json` and use `System.Text.Json` for all purposes.

PR #110, opened 2023-01-22
GameServer improvements

timiimit:
- Added service handling wait time estimation to join QuickPlay match
- Fix GameServer filtering for QP matches
- Add full support for allowing QP matches to connect
- Add models which work towards handling #33 (switching towards System.Text.Json)


UT4MasterServer/Services/MatchmakingService.cs
```diff
@@ -26,6 +26,14 @@ public async Task<bool> DoesExistWithSessionAsync(EpicID sessionID)
 		return count > 0;
 	}
 
+	public async Task<bool> DoesExistAsync(EpicID serverID)
```
belmirp, line 29: Maybe ExistsAsync.

UT4MasterServer/Controllers/UnrealTournamentWaitTimesController.cs
```diff
@@ -16,31 +16,32 @@ namespace UT4MasterServer.Controllers;
 [Produces("application/json")]
 public class UnrealTournamentWaitTimesController : JsonAPIController
```
belmirp, line 17: sealed

UT4MasterServer/Models/Requests/WaitTimeEstimate.cs
```diff
@@ -0,0 +1,22 @@
...
+	[JsonPropertyName("ratingType")]
+	public string RatingType { get; set; }
+
+	[JsonPropertyName("averageWaitTimeSecs")]
+	public double WaitTimeSeconds { get; set; }
+
+	[JsonPropertyName("numSamples")]
+	public int SampleCount { get; set; }
+
+	public WaitTimeEsitmate(string ratingType, double seconds, int sampleCount)
```
belmirp, line 16: EstimateWaitTime

UT4MasterServer/Services/MatchmakingService.cs
```diff
@@ -72,31 +80,13 @@ public async Task<List<GameServer>> ListAsync(GameServerFilter inputFilter)
 		if (inputFilter.BuildUniqueId != null)
 			doc.Add(new BsonElement(nameof(GameServer.BuildUniqueID), inputFilter.BuildUniqueId));
```
belmirp, line 81: You should avoid adding statement without braces in C#. Statement block without braces is more of JS thing.
You could leave only short statement without braces, like return/continue/break.

UT4MasterServer/Services/MatchmakingWaitTimeEstimateService.cs
```diff
@@ -0,0 +1,55 @@
+using UT4MasterServer.Models.Requests;
+
+namespace UT4MasterServer.Services;
+
+public class MatchmakingWaitTimeEstimateService
```
belmirp, line 5: sealed

UT4MasterServer/Services/MatchmakingWaitTimeEstimateService.cs
```diff
@@ -0,0 +1,55 @@
...
+
+	public MatchmakingWaitTimeEstimateService()
+	{
+		estimates = new Dictionary<string, List<(DateTime, double)>>();
+	}
+
+	public void AddWaitTime(string mode, double seconds)
+	{
+		if (!estimates.ContainsKey(mode))
+			estimates.Add(mode, new List<(DateTime, double)>());
```
belmirp, line 18: This is the reason why you should put braces after IFs, this code looks like it is unformatted, instead of being IF block + additional non-related line.

UT4MasterServer/Services/MatchmakingService.cs
```diff
@@ -62,8 +70,8 @@ public async Task<List<GameServer>> ListAsync(GameServerFilter inputFilter)
 		// Build BsonDocument representing Find filter
 		var doc = new BsonDocument();
 
-		// include GameServers that have started
-		doc.Add(new BsonElement(nameof(GameServer.Started), true));
+		//// include GameServers that have started
```
Saibamen, line 73:
```suggestion
		// include GameServers that have started
```

timiimit: There are 4 slashes because i re-commented this and commented one line under this. It's kind of indicating that this was a comment to the code in commented next line.

UT4MasterServer/Models/Requests/WaitTimeEstimate.cs
timiimit: I named it `WaitTimeEstimate` because this is returned from endpoint `.../wait_time/estimate`.

Even if that wasn't the case, for some reason it sounds better to me this way too. I guess i think differently.

belmirp: The endpoint is structured differently, it has **wait_time** as location, and **estimate** as method name.

Convention is that method names are always verbs, not nouns.

Also, you had a typo in Estimate.

timiimit: > Convention is that method names are always verbs, not nouns.

But this is the name of a class and its constructor, not a normal method.

UT4MasterServer/Models/Requests/WaitTimeEstimateResponse.cs
```diff
@@ -13,7 +13,7 @@ public sealed class WaitTimeEsitmate
 	[JsonPropertyName("numSamples")]
 	public int SampleCount { get; set; }
 
-	public WaitTimeEsitmate(string ratingType, double seconds, int sampleCount)
+	public WaitTimeEstimateResponse(string ratingType, double seconds, int sampleCount)
```
timiimit, line 16: I renamed class to `WaitTimeEstimateResponse`, does this sound alright?

UT4MasterServer/Models/Requests/WaitTimeEstimate.cs
belmirp: I thought it was a method, didn't see it was a constructor. Then it is OK. Just fix the typo.

UT4MasterServer/Models/Requests/WaitTimeEstimateResponse.cs
belmirp: It sounds OK. But you should move it to Responses folder instead of Requests.

belmirp (approved):

PR #114, opened 2023-01-22
Optimize images

Saibamen:
### Flags
image

### Avatars
image


timiimit:
Uruguay and Zimbabwe increased in size.
I'm not sure if we want to compress these images. Images already are small and if we compress them we lose originals.
Also i maybe see some slight differences between some images. Maybe only compressing non-complex images would be better.
idk, i have to think about this.

Saibamen:
> Uruguay and Zimbabwe increased in size.
In screenshot, they are this same

timiimit:
github says they are larger.

image

Saibamen:
Fixed

timiimit:
OK i have decided. Please remove changes of avatars, i would like to just optimize Flags.

Saibamen:
Done

timiimit (approved):

PR #127, opened 2023-01-23
Frontend: Show Hub IP

Saibamen:
And enable user select to copy it

image

image

### DIFF
--- a/UT4MasterServer.Web/src/components/Hub.vue
+++ b/UT4MasterServer.Web/src/components/Hub.vue
@@ -3,6 +3,7 @@
         :title="hub?.attributes.UT_SERVERNAME_s" :class="{ 'active-hub': selectedHub?.id === hub?.id }">
         <template v-if="hub">
             <h5>{{ hub.attributes.UT_SERVERNAME_s }}</h5>
+            <div>IP: <span class="hub-ip">{{ hub.serverAddress }}</span></div>
             <div class="flex-space-btw">
                 <div>{{ hub.matches.length }} Matches</div>
                 <div>{{ totalPlayers }} Players</div>
@@ -27,6 +28,10 @@
         text-overflow: ellipsis;
         overflow: hidden;
         white-space: nowrap;
+    }
+
+    span.hub-ip {
+        user-select: all;
     }
 }
 </style>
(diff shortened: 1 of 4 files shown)


UT4MasterServer.Web/src/components/Hub.vue
```diff
@@ -3,6 +3,7 @@
         :title="hub?.attributes.UT_SERVERNAME_s" :class="{ 'active-hub': selectedHub?.id === hub?.id }">
         <template v-if="hub">
             <h5>{{ hub.attributes.UT_SERVERNAME_s }}</h5>
+            <div>IP: <span class="hub-ip">{{ hub.serverAddress }}</span></div>
```
caseyswilliams, line 6: If you use the <CodeBlock> component you'd get the handy little copy button for free

caseyswilliams (approved):

UT4MasterServer.Web/src/components/Hub.vue
Saibamen: It doesn't look very good IMO

image

caseyswilliams: Eh, you can leave it as is, I'll add the copy button in later

Saibamen: Without `IP: `:

image

caseyswilliams: That looks better than than with the IP label, you can go either way, your call

timiimit: I think it looks weird having whole empty line just for COPY on the other side. Not having any extra button and just clicking directly onto the IP could copy it.

caseyswilliams: That's easy enough, @Saibamen you can just leave it as your original and I'll add the click handler to copy if you want

Saibamen: IMO adding **auto** click&copy handler will be bad (without any text info ("Click to copy"), because user will not know about updating his clipboard.

Current version is in my first PR comment. Ready for merge

timiimit (approved):

PR #129, opened 2023-01-23
Improve security

timiimit:
- Fix password change vulnrability
- Change responses on account edits
- Remove try-catch and let global exception handler handle exceptions

### DIFF
--- a/UT4MasterServer/Services/AccountService.cs
+++ b/UT4MasterServer/Services/AccountService.cs
@@ -87,6 +87,5 @@ public sealed class AccountService
 	{
 		await accountCollection.DeleteOneAsync(user => user.ID == id);
 	}
-
-
 }
+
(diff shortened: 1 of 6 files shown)


UT4MasterServer/Services/AccountService.cs
```diff
@@ -33,95 +32,61 @@ public async Task CreateAccountAsync(string username, string email, string passw
...
+		await accountCollection.ReplaceOneAsync(user => user.ID == updatedAccount.ID, updatedAccount);
+	}
+
+	public async Task RemoveAccountAsync(EpicID id)
+	{
+		await accountCollection.DeleteOneAsync(user => user.ID == id);
+	}
+
+
+}
```
Saibamen, line 92:
```suggestion
	}
}

```

PR #145, opened 2023-01-25
Fix unused codes wasting memory

timiimit:
- Remove old unused code if user generates a new one
- Remove unused and expired codes in a background task

### DIFF
--- a/UT4MasterServer/Services/ApplicationBackgroundService.cs
+++ b/UT4MasterServer/Services/ApplicationBackgroundService.cs
@@ -51,6 +51,11 @@ public sealed class ApplicationBackgroundCleanupService : IHostedService, IDispo
 			if (deleteCount > 0)
 				logger.LogInformation("Background task deleted {DeleteCount} expired sessions.", deleteCount);
 
+			var codeService = scope.ServiceProvider.GetRequiredService<CodeService>();
+			deleteCount = await codeService.RemoveAllExpiredCodesAsync();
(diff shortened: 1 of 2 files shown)


Saibamen (approved):

UT4MasterServer/Services/ApplicationBackgroundService.cs
```diff
@@ -51,6 +51,11 @@ private void DoWork(object? state)
 			if (deleteCount > 0)
 				logger.LogInformation("Background task deleted {DeleteCount} expired sessions.", deleteCount);
 
+			var codeService = scope.ServiceProvider.GetRequiredService<CodeService>();
+			deleteCount = await codeService.RemoveAllExpiredCodesAsync();
+			if (deleteCount > 0)
+				logger.LogInformation("Background task deleted {DeleteCount} expired codes.", deleteCount);
```
belmirp, line 57: If you are trying to save some space by adding condition on the log message, why just not remove it?
If you aren't trying to save some space, just remove IF.

belmirp (approved):

UT4MasterServer/Services/ApplicationBackgroundService.cs
timiimit: i am trying to remove clutter in console and there's no reason to say when nothing was deleted. I was at some point annoyed because it was spitting out 2 lines of practically no info in debug every minute .

PR #146, opened 2023-01-25
Convert emails to lowercase before any further processing

timiimit:
Previously registration and change email operations failed due to regex validation only allowing lowercase

### DIFF
--- a/UT4MasterServer/Controllers/AccountController.cs
+++ b/UT4MasterServer/Controllers/AccountController.cs
@@ -250,6 +250,7 @@ public sealed class AccountController : JsonAPIController
 			return Conflict("You have entered an invalid username");
 		}
 
+		email = email.ToLower();
 		account = await accountService.GetAccountByEmailAsync(email);
 		if (account != null)
 		{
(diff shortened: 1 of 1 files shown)


belmirp (approved):

UT4MasterServer/Controllers/AccountController.cs
```diff
@@ -250,6 +250,7 @@ public async Task<IActionResult> RegisterAccount([FromForm] string username, [Fr
 			return Conflict("You have entered an invalid username");
 		}
 
+		email = email.ToLower();
```
Saibamen, line 253:
```suggestion
		email = email?.ToLower();

```

UT4MasterServer/Controllers/AccountController.cs
```diff
@@ -325,6 +326,7 @@ public async Task<IActionResult> UpdateEmail([FromForm] string newEmail)
 			return Unauthorized();
 		}
 
+		newEmail = newEmail.ToLower();
```
Saibamen, line 329:
```suggestion
		newEmail = newEmail?.ToLower();

```

Saibamen (changes requested):
Prevent NRE

UT4MasterServer/Controllers/AccountController.cs
timiimit: asp.net core makes sure the value is not null. if it is, it will not call the method as far as i know.

Saibamen (approved):

PR #153, opened 2023-01-29
Servers page mobile optimization, filtering, styling

caseyswilliams:
- Changed servers page structure so it will format better on mobile/smaller viewports
- Added filtering for hubs, filter out empty hubs by default
- Sorted hubs by server trust level asc (more trust first), totalPlayers desc
- Added visual identification of trusted servers
- Removed Server IP
- Formatted time elapsed in match
- Added font awesome icons
- Updated copy button to use icons
- Moved bootstrap css to public so we can have a little style in the news pages


UT4MasterServer/Controllers/PersonaController.cs
```diff
@@ -75,8 +75,8 @@ public async Task<IActionResult> GetAllAccounts()
 		foreach (var account in accounts.Take(1000))
 		{
 			var obj = new JObject();
-			obj.Add("id", account.ID.ToString());
-			obj.Add("displayName", account.Username);
+			obj.Add("ID", account.ID.ToString());
+			obj.Add("Username", account.Username);
```
Saibamen, line 79: image

caseyswilliams: I'm not sure what you're trying to say. The endpoints in the account controller are used by the game, and thus must use that json format. The endpoints in persona controller were created solely for the front end and are returning the Account model as is. We did this because we didn't want to modify the json format the game uses, but the front end needed access to the flags/roles

Saibamen: If it is for FE only, we're good

Saibamen: Can you add info about it to `<summary>` tag for this controller?

timiimit (approved):

PR #154, opened 2023-01-30
Admin panel - Clients and Trusted Servers

caseyswilliams:
- Added Clients and Trusted Servers CRUD screens to Admin
- Added change password and delete account
- Fixed bug in set account flags, logic was inverted


UT4MasterServer.Web/src/pages/Admin/Clients/components/AddClient.vue
```diff
@@ -0,0 +1,64 @@
+<template>
+    <LoadingPanel :status="status" :error="errorMessage">
+        <form :class="{ 'was-validated': submitAttempted }" @submit.prevent="handleSubmit" novalidate>
+            <fieldset>
+                <legend>Add Client</legend>
+                <div class="form-group row">
+                    <label for="name" class="col-sm-12 col-form-label">Name</label>
+                    <div class="col-sm-6">
+                        <input type="text" class="form-control" id="name" name="name" v-model="name" v-valid="formValid" required />
+                        <div class="invalid-feedback">A unique name is required</div>
```
timiimit, line 10: Name is not really required for backed to work. It is meant as just some description for visual display.

UT4MasterServer.Web/src/pages/Admin/Clients/hooks/use-client-options.hook.ts
```diff
@@ -0,0 +1,24 @@
+import { ITrustedGameServer } from "../../TrustedServers/types/trusted-game-server";
+import { IClient } from "../types/client";
+
+export function useClientOptions() {
+    const reservedNames = ['Launcher (our website)', 'Game', 'ServerInstance'];
```
timiimit, line 5: Names are not required to be set.
If there's a reason for you to check whether client is reserved or not, it should check the client id and not the name.

UT4MasterServer.Web/src/pages/Admin/Accounts/Accounts.vue
```diff
@@ -75,4 +90,29 @@ async function loadAccounts() {
     }
 }
 
+function canDelete(account: IGridAccount) {
+    return AccountStore.account?.ID !== account.ID && !account.Roles?.includes(AccountFlag.Admin);
+}
```
timiimit, line 95: Similar as for clients. I don't think its really up to the frontend to check whether the operation is possible. Backend might change and then frontend will still use old conditions. Which means that both frontend and backend need to be changed if any conditions might need to change.

I can already tell you that this condition is not equivalent to the condition that backend checks. Moderators are also capable of deleting accounts and no one is capable of deleting account on the same or higher level of status (in order: admin, moderator, others).

UT4MasterServer.Web/src/pages/Admin/Accounts/Accounts.vue
```diff
@@ -12,35 +8,45 @@
                 <thead>
                     <tr>
                         <th>Username</th>
+                        <th>Roles</th>
                         <th />
```
timiimit, line 12: Is this also intentional?

UT4MasterServer.Web/src/pages/Admin/Clients/Clients.vue
```diff
@@ -0,0 +1,119 @@
...
+                        <th>Name</th>
+                        <th>ID</th>
+                        <th />
+                    </tr>
+                </thead>
+                <tbody>
+                    <template v-for="client in filteredClients.slice(pageStart, pageEnd)" :key="objectHash(client)">
+                        <tr :class="{ 'table-light': client.editing }">
+                            <td>{{ client.name }}</td>
+                            <td>{{ client.id }}</td>
```
timiimit, line 22: I would swap these two
```suggestion
                            <td>{{ client.id }}</td>
                            <td>{{ client.name }}</td>
```

UT4MasterServer.Web/src/pages/Admin/Clients/types/client.ts
```diff
@@ -0,0 +1,5 @@
+export interface IClient {
+    id: string;
+    secret: string;
+    name: string;
```
timiimit, line 4: name is optional on backend

UT4MasterServer/Controllers/AdminPanelController.cs
```diff
@@ -111,11 +111,11 @@ public async Task<IActionResult> SetAccountFlags(string accountID, [FromBody] st
 
 
 	[HttpPost("clients/new")]
-	public async Task<IActionResult> CreateClient()
+	public async Task<IActionResult> CreateClient([FromBody]string name)
```
timiimit, line 114: Should be optional
```suggestion
	public async Task<IActionResult> CreateClient([FromBody] string? name)
```

timiimit (changes requested):

UT4MasterServer.Web/src/pages/Admin/Clients/components/AddClient.vue
caseyswilliams: Yeah I know it isn't required in the backend, but without it you would have a dropdown with 32 digit IDs in it, not the easiest to use. I made it required because it would be a horrible pain to use if you didn't set a name.

timiimit (approved):

PR #158, opened 2023-02-01
Refactor all accounts

caseyswilliams:
As per a discussion with @timiimit, refactored the front end to remove unbound calls get accounts in favor of querying/paging. This turned out to be a bit more involved than I thought it would be, but it is much more efficient now and should use considerably less bandwidth. Hopefully I didn't introduce any weird bugs, but I think I tested pretty thoroughly.


UT4MasterServer/Services/AccountService.cs
```diff
@@ -46,6 +46,12 @@ public async Task CreateAccountAsync(string username, string email, string passw
 		return await cursor.SingleOrDefaultAsync();
 	}
 
+	public async Task<IEnumerable<Account>> SearchAccountsAsync(string query)
+	{
+		var cursor = await accountCollection.FindAsync(account => account.Username.ToLower().Contains(query.ToLower()));
```
timiimit, line 51: Did you made sure searching like this works? i would think it doesn't since you do "so many" operations on the strings.

caseyswilliams: Yep it works. Mongo supports toLower, but it doesn't support Contains(str, StringComparison.ignoreCase)

UT4MasterServer/Controllers/PersonaController.cs
```diff
@@ -32,58 +33,91 @@ public async Task<IActionResult> AccountLookup([FromQuery(Name = "q")] string qu
...
+	public async Task<IActionResult> GetAccountsByIds([FromBody] string[] ids)
+	{
+		if (User.Identity is not EpicUserIdentity authenticatedUser)
+		{
+			return Unauthorized();
+		}
+
+		var eIds = ids.Distinct().Select(x => EpicID.FromString(x));
+		var accounts = await accountService.GetAccountsAsync(eIds);
+		logger.LogInformation($"{authenticatedUser.Session.AccountID} is looking for limited accounts by ID");
```
belmirp, line 119: Don't use string interpolation in logging, use this:
```charp
logger.LogInformation("{AccountId} is looking for limited accounts by ID", authenticatedUser.Session.AccountID);
```

UT4MasterServer/Models/Requests/AccountSearchRequest.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System.Text.Json.Serialization;
+
+namespace UT4MasterServer.Models.Requests;
+
+public class AccountSearchRequest
```
belmirp, line 5: sealed

UT4MasterServer/Models/Responses/TrustedGameServerResponse.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace UT4MasterServer.Models.Responses;
+
+public class TrustedGameServerResponse: TrustedGameServer
```
belmirp, line 3: sealed

UT4MasterServer/Controllers/PersonaController.cs
caseyswilliams: I just copied this from another function, so i don't mind changing it, but curious how that syntax would even work? Are we parsing the strings going into the logger or something?

belmirp: When logs are exported to something like Kibana/ElasticSearch, {AccountId} becomes a variable that you can use to search by.

caseyswilliams: So is it a specific set of variables that we have access to in the context of a logged method invocation? Like if I wanted to log something specific to that occurrence, say the result of a calculation or something that only that invocation would have access to, I'd still need to interpolate that, right?

belmirp: No, the variable is whatever you put in {}. Once the logs are dumped to a service like ElasticSearch, the service scans all logs and extracts all variables so you can use them in filters. When using these variables, you should be consistent with naming and data type. If you log once a string as {AccountId}, you should stick with it, and don't change it to int or something else.

Basic data types are supported like bools, numbers, strings and dates. For example, for dates you could use filter between or greaten than, etc.

caseyswilliams: Oh, duh, I didn't see that you were passing the ID as a second param, that's why I was confused. That's what I get for looking at comments on my phone.

belmirp (approved):

PR #159, opened 2023-02-01
GH Actions: Check linters. Fix 1 deprecated warning

Saibamen:


UT4MasterServer.Web/package.json
```diff
@@ -8,7 +8,8 @@
     "dev-open": "vite --host 0.0.0.0 --open",
     "build": "vite build",
     "build-dev": "vite build --mode development",
-    "format": "prettier . -w",
+    "format": "prettier . --write",
```
caseyswilliams, line 11: These are equivalent

.github/workflows/linters.yml
```diff
@@ -0,0 +1,33 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: lts/*
+      - run: npm install --ignore-scripts
+        working-directory: ${{ env.WEB_DIR }}
+      - run: npm run lint
+        working-directory: ${{ env.WEB_DIR }}
+      - run: npm run format-check
```
caseyswilliams, line 32: This isn't strictly necessary because I have prettier write and eslint fix running as a commit hook, but I wouldn't be against making the github action do prettier write and lint fix and removing the husky hook

caseyswilliams (approved):
Fine as is, maybe a bit of overkill 🤷

.github/workflows/linters.yml
Saibamen: I won't recommend making commits fixes from auto-linters directly from GH Actions.
They can break your code, for example like auto fixes for `prefer-template` rule (more than 2 spaces, sometimes no spaces)

PR #161, opened 2023-02-02
Fix gameserver player updates

timiimit:


UT4MasterServer/Controllers/UnrealTournamentMatchmakingController.cs
```diff
@@ -235,8 +235,29 @@ public async Task<IActionResult> UpdateGameServerPlayers(string id, [FromBody] G
...
-		server.PublicPlayers = serverOnlyWithPlayers.PublicPlayers;
-		server.PrivatePlayers = serverOnlyWithPlayers.PrivatePlayers;
+		// handle player list update
+		foreach (var player in serverOnlyWithPlayers.PublicPlayers)
+		{
+			if (!server.PublicPlayers.Where(x => x == player).Any())
+			{
+				server.PublicPlayers.Add(player);
+			}
+			if (server.PrivatePlayers.Where(x => x == player).Any())
```
belmirp, line 245: else

UT4MasterServer/Controllers/UnrealTournamentMatchmakingController.cs
```diff
@@ -235,8 +235,29 @@ public async Task<IActionResult> UpdateGameServerPlayers(string id, [FromBody] G
...
+				server.PrivatePlayers.Remove(player);
+			}
+		}
+		foreach (var player in serverOnlyWithPlayers.PrivatePlayers)
+		{
+			if (!server.PrivatePlayers.Where(x => x == player).Any())
+			{
+				server.PrivatePlayers.Add(player);
+			}
+			if (server.PublicPlayers.Where(x => x == player).Any())
```
belmirp, line 256: else

UT4MasterServer/Controllers/UnrealTournamentMatchmakingController.cs
```diff
@@ -258,8 +279,10 @@ public async Task<IActionResult> RemovePlayer(string id, [FromBody] EpicID[] pla
 
 		foreach (var player in players)
 		{
-			server.PublicPlayers.Remove(player);
-			server.PrivatePlayers.Remove(player);
+			if (!server.PublicPlayers.Remove(player))
```
belmirp, line 282: What is server.PublicPlayers.Remove supposed to return? Bool result? Or you meant Any()?

UT4MasterServer/Controllers/UnrealTournamentMatchmakingController.cs
timiimit: The way i wrote it is the intended logic. I want to add it to `PublicPlayers` if it doesn't exist there yet and i want to remove it from `PrivatePlayers` if it exists there. This way we ensure that there is only one instance of some player across both arrays.

UT4MasterServer/Controllers/UnrealTournamentMatchmakingController.cs
timiimit: same as above, but now reverse role of `PrivatePlayers` and `PublicPlayers`

UT4MasterServer/Controllers/UnrealTournamentMatchmakingController.cs
timiimit: it returns `true` if it was removed, otherwise `false`. in case player was removed from `PublicPlayers` then it shouldn't also exist in `PrivatePlayers`.

UT4MasterServer/Controllers/UnrealTournamentMatchmakingController.cs
belmirp: On the first sight, it looked to me that they were both same list.

belmirp (approved):

Issue #138, opened 2023-01-25
Handle Ratings


timiimit:
Each player needs to have an associated ELO rating which is adjusted based on their performance in matches. All of this should be implemented in `UnrealTournamentStatsController`.

PR #162, opened 2023-02-02
Elo implementation

belmirp:
Implementation for ELO rating system for both deathmatch and team-based game modes.

The attached Excel file contains calculation for testing.

[ELO Calculation.xlsx](https://github.com/timiimit/UT4MasterServer/files/10572676/ELO.Calculation.xlsx)

Related: [#138](https://github.com/timiimit/UT4MasterServer/issues/138)


UT4MasterServer/Services/RatingsService.cs
```diff
@@ -0,0 +1,260 @@
...
+					result.Ratings.Add(ratings.DMSkillRating / Rating.Precision);
+					result.NumGamesPlayed.Add(ratings.DMSkillRatingGamesPlayed);
+					break;
+			}
+		}
+
+		return result;
+	}
+
+	public async Task<JObject> GetRatingAsync(EpicID accountID, string ratingType)
```
timiimit, line 71: This is just a preference comment, you don't have to do this if you don't feel like it.

We are trying to go away from Newtonsoft (#33) so creating a class for the returned object would be great.

timiimit (changes requested):
There aren't just SkillRating, TDMSkillRating, CTFSkillRating, ShowdownSkillRating, FlagRunSkillRating, DMSkillRating.

These are custom strings which can be possibly be modified by master server to some extent. I think that `Rating` class should contain an array of `ModeRating` type which you removed from MMRBulk.cs file. But the "SkillRating" substring at the end of each type doesn't need to be saved since it is there at all times.

The elo calculation static classes arent really services, placing them in Helpers folder seems more appropriate.

belmirp:
> 

There is an example of POST request where ratingType=SkillRating in those Fiddler sessions that you provided (see file ut4-gameplay-local-hub (1)).

I saw `ModeRating`, and I thought it was just a suggestion how to store data. I could make everything to be dynamic as you suggested and save whatever string as `RatingType` is sent in the request. There is however one problems with that. ELO calculation for deathmatch is different from team-based mode, and I need to know which game mode is, and that information is missing from the request. Maybe we could use the data that is sent in `RatingMatch` object to determine if game is deathmatch or not, e.g. `blueTeam.member` is an empty array.

In my opinion, we should accept only those strings that are available now in the game client. If you say that you can customize these strings with Master Server, then once you do it, we should go back and expand the logic to accept more rating types.

UT4MasterServer/Models/Rating.cs
```diff
@@ -0,0 +1,52 @@
...
+public sealed class Rating
+{
+	public const int DefaultRating = 1500;
+	public const int Precision = 1000;
+
+	[BsonId, BsonIgnoreIfDefault, BsonElement("_id"), BsonRepresentation(BsonType.ObjectId)]
+	public string ID { get; set; } = default!;
+
+	public EpicID AccountID { get; set; } = default!;
+
```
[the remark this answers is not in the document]
belmirp: I will do that, since every account should have only one record related to ratings.

UT4MasterServer/Controllers/UnrealTournamentRatingsController.cs
```diff
@@ -72,18 +72,48 @@ public IActionResult LeagueRating(string id, string leagueName)
...
 
 	[HttpPost("team/match_result")]
-	public IActionResult MatchResult([FromBody] RatingMatch body)
+	public async Task<IActionResult> MatchResult([FromBody] RatingMatch ratingMatch)
 	{
-		// TODO: update ELO rating
+		switch (ratingMatch.RatingType)
+		{
+			case "SkillRating":
+				await ratingsService.UpdateTeamsRatingsAsync(ratingMatch, x => x.SkillRating);
```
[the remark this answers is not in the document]
belmirp: Here is an example of POST that has ratingType=SkillRating:

```json
{
	"ratingType": "SkillRating",
	"matchInfo":
	{
		"sessionId": "a649683b6b494e179ab57129ba23b743",
		"redScore": 1,
		"matchLengthSeconds": 705
	},
	"redTeam":
	{
		"members": [
			{
				"accountId": "0b0f09b400854b9b98932dd9e5abe7c5",
				"score": 0,
				"isBot": false
			}
		],
		"socialPartySize": 1
	},
	"blueTeam":
	{
		"members": [
			{
				"accountId": "ebae35e47c64455fa63625052d6f37ee",
				"score": 0,
				"isBot": false
			}
		],
		"socialPartySize": 1
	}
}
```
It looks more like a duel to me.

UT4MasterServer/Services/RatingsService.cs
belmirp: I will remove JObject, I just followed the existing code.

timiimit:
It does make sense to only allow values that we know. That way malicious clients cant do as much harm. Alright, its fine to leave it statically defined. But can you please find a way to get rid of repetitions in switches?

belmirp:
> It does make sense to only allow values that we know. That way malicious clients cant do as much harm. Alright, its fine to leave it statically defined. But can you please find a way to get rid of repetitions in switches?

I will check that, but I am not optimistic, since these are always 6 different properties, except if you want to go with the reflection.

timiimit:
> > It does make sense to only allow values that we know. That way malicious clients cant do as much harm. Alright, its fine to leave it statically defined. But can you please find a way to get rid of repetitions in switches?
> 
> I will check that, but I am not optimistic, since these are always 6 different properties, except if you want to go with the reflection.

You can make a dictionary of mappings from string to property selectors

UT4MasterServer.Services/Scoped/RatingsService.cs
```diff
@@ -0,0 +1,210 @@
+﻿using MongoDB.Driver;
+using UT4MasterServer.Common;
+using UT4MasterServer.Models;
+using UT4MasterServer.Models.DTO.Response;
+using UT4MasterServer.Models.DTO.Responses;
+
+namespace UT4MasterServer.Services;
```
timiimit, line 7: Use right namespace based on file location
```suggestion
namespace UT4MasterServer.Services.Scoped;
```

UT4MasterServer.Models/Database/Rating.cs
```diff
@@ -0,0 +1,41 @@
+﻿using MongoDB.Bson.Serialization.Attributes;
+using UT4MasterServer.Common;
+
+namespace UT4MasterServer.Models;
```
timiimit, line 4:
```suggestion
namespace UT4MasterServer.Models.Database;
```

UT4MasterServer.Models/DTO/Response/MMRBulkResponse.cs
```diff
@@ -2,7 +2,7 @@
 
 namespace UT4MasterServer.Models.DTO.Responses;
 
-public sealed class MMRBulk
+public class MMRBulkResponse
```
timiimit, line 5: Missing sealed

UT4MasterServer.Models/Database/Rating.cs
```diff
@@ -0,0 +1,41 @@
...
+		"RankedDuelSkillRating",
+		"RankedCTFSkillRating",
+		"RankedShowdownSkillRating",
+		"RankedFlagRunSkillRating",
+	};
+
+	public static readonly List<string> AllowedRatingTypes = DmRatingTypes.Union(TeamRatingTypes).ToList();
+
+	[BsonId]
+	public EpicID AccountID { get; set; } = default!;
```
timiimit, line 32: How can AccountID have [BsonId] when you now have one entry per rating type?

UT4MasterServer/Controllers/UT/RatingsController.cs
```diff
@@ -1,10 +1,9 @@
 ﻿using Microsoft.AspNetCore.Mvc;
-using Microsoft.Extensions.Hosting;
-using Newtonsoft.Json.Linq;
 using UT4MasterServer.Authentication;
-using UT4MasterServer.Controllers.Epic;
+using UT4MasterServer.Common;
+using UT4MasterServer.Models;
 using UT4MasterServer.Models.DTO.Responses;
-using UT4MasterServer.Services.Scoped;
+using UT4MasterServer.Services;
```
timiimit, line 6: Keep .Scoped as the service's should be there.

UT4MasterServer.Services/Scoped/RatingsService.cs
```diff
@@ -0,0 +1,210 @@
...
+	private readonly IMongoCollection<Rating> ratingsCollection;
+
+	public RatingsService(DatabaseContext dbContext)
+	{
+		ratingsCollection = dbContext.Database.GetCollection<Rating>("ratings");
+	}
+
+	public async Task<MMRBulkResponse> GetRatingsAsync(EpicID accountID, MMRBulkResponse mmrBulk)
+	{
+		var ratingTypes = mmrBulk.RatingTypes.Intersect(Rating.AllowedRatingTypes);
```
timiimit, line 20: If a request for some rating we don't support have comes in, we still want to return a default rating for specified rating type.

UT4MasterServer.Services/Scoped/RatingsService.cs
```diff
@@ -0,0 +1,210 @@
...
+	{
+		var filter = Builders<Rating>.Filter.Eq(f => f.AccountID, accountID) &
+					 Builders<Rating>.Filter.Eq(f => f.RatingType, ratingType);
+		var rating = await ratingsCollection.Find(filter).FirstOrDefaultAsync();
+
+		var result = new MMRRatingResponse()
+		{
+			Rating = rating?.RatingValue / Rating.Precision ?? Rating.DefaultRating,
+			GamesPlayed = rating?.GamesPlayed ?? 0,
+		};
```
timiimit, line 49: do
```cs
if (rating == null)
{
	return new MMRRatingResponse();
}
```
so you don't have to do null checks later

UT4MasterServer.Models/DTO/Response/MMRRatingResponse.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System.Text.Json.Serialization;
+
+namespace UT4MasterServer.Models.DTO.Response;
+
+public sealed class MMRRatingResponse
+{
+	[JsonPropertyName("rating")]
+	public int Rating { get; set; }
```
timiimit, line 8: give a default value 1500

UT4MasterServer.Models/DTO/Response/MMRRatingResponse.cs
```diff
@@ -0,0 +1,12 @@
...
+
+	[JsonPropertyName("numGamesPlayed")]
+	public int GamesPlayed { get; set; }
```
timiimit, line 11: give explicit default value 0

UT4MasterServer.Services/Scoped/RatingsService.cs
```diff
@@ -0,0 +1,210 @@
...
+
+		var result = new MMRBulkResponse();
+
+		foreach (var ratingType in ratingTypes)
+		{
+			var rating = ratings.Where(w => w.RatingType == ratingType).FirstOrDefault();
+
+			result.RatingTypes.Add(ratingType);
+			result.Ratings.Add(rating?.RatingValue / Rating.Precision ?? Rating.DefaultRating);
+			result.NumGamesPlayed.Add(rating?.GamesPlayed ?? 0);
```
timiimit, line 33: check for null first

UT4MasterServer.Services/Scoped/RatingsService.cs
```diff
@@ -0,0 +1,210 @@
...
+			.ToArray();
+		var filter = Builders<Rating>.Filter.In(f => f.AccountID, accountIds) &
+					 Builders<Rating>.Filter.Eq(f => f.RatingType, ratingType);
+		var ratings = await ratingsCollection.Find(filter).ToListAsync();
+
+		List<int> ratingValues = new();
+
+		foreach (var member in ratingTeam.Members.Where(w => !w.IsBot))
+		{
+			var ratingValue = ratings.Where(w => w.AccountID == EpicID.FromString(member.AccountID)).FirstOrDefault()?.RatingValue / Rating.Precision ?? Rating.DefaultRating;
```
timiimit, line 68: split LINQ into multiple lines to avoid long line

timiimit (changes requested):
You made RatingType change value instead of updating the existing stored RatingType?

UT4MasterServer.Services/Scoped/RatingsService.cs
```diff
@@ -0,0 +1,210 @@
...
+			.Select(s => EpicID.FromString(s.AccountID))
+			.ToArray();
+
+		int playersCount = playersAccountIds.Length;
+
+		var filter = Builders<Rating>.Filter.In(f => f.AccountID, playersAccountIds);
+
+		var playersCurrentRatings = await ratingsCollection.Find(filter).ToListAsync();
+
+		double[] currentRatings = new double[playersCount];
```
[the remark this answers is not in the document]
belmirp: Done.

UT4MasterServer.Models/Database/Rating.cs
belmirp: Done.

UT4MasterServer.Services/Scoped/RatingsService.cs
belmirp: Done.

UT4MasterServer.Models/DTO/Response/MMRBulkResponse.cs
belmirp: Done.

UT4MasterServer.Models/DTO/Response/MMRRatingResponse.cs
belmirp: I did that in the constructor.

UT4MasterServer.Models/DTO/Response/MMRRatingResponse.cs
belmirp: Int values are 0 by default.

belmirp: Did that in the constructor of the response object.

UT4MasterServer.Services/Scoped/RatingsService.cs
belmirp: Done.

timiimit (approved):

Issue #26, opened 2022-12-29
Add extended support for friend list


timiimit:
XMPP server is used to make chat and user persence possible within the game. We need to set-up, configure and connect some XMPP server to our mongodb.

User address for logging in should be `<accountID>@prod.ol.epicgames.com/V2:UnrealTournamentDev:::<randomGUID>` and password should be access token of currently logged in user.

For initial testing an external XMPP client can be used. One that i know that it can connect to existing EpicGames server is [Swift](https://swift.im/).

PR #164, opened 2023-02-03
Organize repo - Major codebase change

timiimit:
This PR organizes repository. I don't expect anyone to look through the code.

The following was performed:
- put related classes into same namespace and folder
- separate the backend code into multiple project modules
- at a couple of places Newtonsoft library was replaced with System.Text.Json which follows #33
- a couple more minor changes were made

c# code was split up because I was working on #26. I wrote a c# XMPP server which needs some knowledge of classes used in main master server and both will need to communicate directly (most likely over a simple tcp connection).

Before merging i still need to test whether everything is operational. When that's done, I will duplicate current master branch to another branch. This way it will be easier to see what this PR messed up if later down the line we discover something isn't working.


timiimit:
Looks like everything is working. I will merge this after #162 is merged.

Saibamen:
I'm will not be available this weekend. Can you check if our paths (both trigger and ignore) in GH Actions need to be changed? Thanks

timiimit:
As agreed with @belmirp I merged this before #162

timiimit:
@Saibamen I think it looks fine as is, but I'm no expert.

PR #165, opened 2023-02-05
Various front-end bugs

caseyswilliams:
- Account admin allowed for changing of Admin and Moderator passwords, fixed this
- 401s resulted in logout, which was fine until it wasn't. Removed this in favor of only logging a user out when the checkAuth call fails.
- I was storing the user's account in local storage, but this caused a potential error scenario if a user never logged out of 1 account and then created another, they would see the original account information. Removed the local storage, this shouldn't result in considerably more calls to getAccount and avoids the error scenario. The account is still stored in memory so it doesn't call getAccount on page load unless the browser is refreshed.

### DIFF
--- a/UT4MasterServer.Web/src/services/http.service.ts
+++ b/UT4MasterServer.Web/src/services/http.service.ts
@@ -1,5 +1,4 @@
 import { SessionStore } from '@/stores/session-store';
-import AuthenticationService from './authentication.service';
 
 type HttpMethod = 'GET' | 'POST' | 'PUT' | 'PATCH' | 'DELETE';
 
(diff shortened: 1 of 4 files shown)


UT4MasterServer.Web/src/services/http.service.ts
```diff
@@ -47,20 +46,12 @@ export default class HttpService {
     const response = await fetch(url, fetchOptions);
 
     if (!response.ok) {
-      // Note: Logging the user out on any API request returning 401 may not be the perfect solution,
-      //       there could be a legitimate case for returning 401 that wouldn't require logout,
-      //       but I don't think we have such a case at this point.
```
caseyswilliams, line 52: Now we do

timiimit (approved):

PR #169, opened 2023-02-05
Cleanup after #164

Saibamen:
Cleanup after #164 and update NuGets to latest versions that support net6.0


timiimit (approved):
i thought i got the latest versions

PR #172, opened 2023-02-06
Account searching improvements

timiimit:
- Create indexes for account username and email
- Optimize db account searching
- Make frontend know that backend port is 80 in dev mode

### DIFF
--- a/UT4MasterServer.Services/Scoped/AccountService.cs
+++ b/UT4MasterServer.Services/Scoped/AccountService.cs
@@ -3,6 +3,7 @@ using MongoDB.Driver;
 using UT4MasterServer.Common;
 using UT4MasterServer.Common.Helpers;
 using UT4MasterServer.Models.Database;
+using UT4MasterServer.Models.DTO.Response;
 using UT4MasterServer.Models.Settings;
 
 namespace UT4MasterServer.Services.Scoped;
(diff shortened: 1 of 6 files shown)


UT4MasterServer.Services/Scoped/AccountService.cs
```diff
@@ -47,10 +59,35 @@ public async Task CreateAccountAsync(string username, string email, string passw
 		return await cursor.SingleOrDefaultAsync();
 	}
 
-	public async Task<IEnumerable<Account>> SearchAccountsAsync(string usernameQuery)
+	public async Task<PagedResponse<Account>> SearchAccountsAsync(string usernameQuery, AccountFlags flagsMask = AccountFlags.All, int skip = 0, int limit = 50)
 	{
-		var cursor = await accountCollection.FindAsync(account => account.Username.ToLower().Contains(usernameQuery.ToLower()));
-		return await cursor.ToListAsync();
+		var f = Builders<Account>.Filter;
```
belmirp, line 64: Unused variable?

UT4MasterServer/Controllers/Epic/PersonaController.cs
```diff
@@ -43,39 +44,59 @@ public async Task<IActionResult> AccountLookup([FromQuery(Name = "q")] string qu
...
+		if (request.Take > 50)
 		{
-			accounts = accounts.Where((a) => a.Roles != null && a.Roles.Intersect(request.Roles).Any());
+			request.Take = 50;
 		}
 
-		var count = accounts.Count();
+		// TODO: duplicate code in AdminPanelController
+		AccountFlags flags;
+		if (request.Roles != null && request.Roles.Length > 0)
```
belmirp, line 64: `request.Roles?.Length` (if possible, not sure).

UT4MasterServer.Services/Scoped/AccountService.cs
timiimit: Odd, visual studio didnt throw any warning about this

UT4MasterServer/Controllers/Epic/PersonaController.cs
timiimit: Even if it is, i don't think its a good idea because it doesnt make it that clear that null check is required.

Saibamen: It's clear and more modern to use conditional access

PR #177, opened 2023-02-07
Cloudstorage improvements

timiimit:
- Rename files with "Cloud**s**torage" to "Cloud**S**torage"
- Rename files with "System**f**iles" to "System**F**iles"
- Fix thrown exception if default system files are missing and unable to be opened
- Fix cloud file deletion bug
- Make system files accessible without authentication to allow announcement website uploads from admin panel


UT4MasterServer/Controllers/AdminPanelController.cs
```diff
@@ -1,422 +1,422 @@
...
+		if (id == ClientIdentification.Game.ID || id == ClientIdentification.ServerInstance.ID || id == ClientIdentification.Launcher.ID)
+		{
+			return true;
+		}
+
+		return false;
+	}
+
+	[NonAction]
+	private async Task<(Session Session, Account Account)> VerifyAdmin()
```
belmirp, line 402: VerifyAdminAsync

belmirp (approved):

PR #178, opened 2023-02-07
Fix line endings

timiimit:
This PR adds .gitattributes to perform line ending normalization and it also normalizes all existing files.


Saibamen (changes requested):
Please also add `end_of_line` to our EditorConfig (docs: https://editorconfig.org/)

Saibamen (approved):

PR #181, opened 2023-02-07
ELO fixes

belmirp:
- Fixed DM ELO calculation
- Excluded accounts with empty account ID from calculation
- Added logs for future monitoring

Related: [#138](https://github.com/timiimit/UT4MasterServer/issues/138)

### DIFF
--- a/UT4MasterServer.Common/Helpers/EloDeathmatchCalculationHelper.cs
+++ b/UT4MasterServer.Common/Helpers/EloDeathmatchCalculationHelper.cs
@@ -35,7 +35,7 @@ public static class EloDeathmatchCalculationHelper
 		for (int i = 0; i < playersCount; i++)
 		{
 			var playersPlace = i + 1;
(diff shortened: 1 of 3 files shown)


Saibamen (changes requested):
We already have timestamp in logs. See our template logs template: https://github.com/timiimit/UT4MasterServer/blob/master/UT4MasterServer/Configuration/InternalLoggerConfiguration.cs#L10

Saibamen (approved):

timiimit (commented):
LogInformation will not be visible in production. I don't think logging every request makes sense. I was expecting you to only log the cases when we detect unexpected content.

Saibamen:
> LogInformation will not be visible in production

For clarity: you can configure to not log info from Microsoft, but log every other or log info for specific controller/services etc

https://github.com/timiimit/UT4MasterServer/blob/master/UT4MasterServer/appsettings.json#L23

PR #188, opened 2023-02-10
Improve AdminPanel

timiimit:
This PR expands AccountFlags to allow more control over who can access what parts of admin panel.
It also forces admins to enter user's email in order to be able to change their password in order to try prevent impersonation.


timiimit:
5th time's the charm

UT4MasterServer.Common/Helpers/AccountFlagsHelper.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using UT4MasterServer.Common.Enums;
+
+namespace UT4MasterServer.Common.Helpers;
+
+public class AccountFlagsHelper
```
belmirp, line 10: static class?

UT4MasterServer.Common/Helpers/EnumHelpers.cs
```diff
@@ -0,0 +1,100 @@
...
+		{
+			var val = allValues[i];
+
+			if (@enum.HasFlag(val))
+			{
+				result.Add(allNames[i]);
+			}
+		}
+
+		return result.ToArray();
```
belmirp, line 34: Why not just return List<T>?

UT4MasterServer.Common/Helpers/EnumHelpers.cs
```diff
@@ -0,0 +1,100 @@
...
+		{
+			var val = allValues[i];
+
+			if (@enum.HasFlag(val))
+			{
+				result.Add(allValues[i]);
+			}
+		}
+
+		return result.ToArray();
```
belmirp, line 88: Same here.

UT4MasterServer.Common/Helpers/EnumHelpers.cs
timiimit: To keep everything in most primitive types.
But since it already is a list it is an unneccessery conversion to plain array operation, i'll make it return a list.

UT4MasterServer.Common/Helpers/AccountFlagsHelper.cs
timiimit: oops

belmirp (approved):

PR #189, opened 2023-02-10
Rankings and Stats UI changes

caseyswilliams:
- Added deep linking capability (if redirected to login when clicking a link, will redirect to original link after login)
- Made Rankings and Stats route params match state of page, so that when changing params the route is updated, allowing for deep linking to any given state as well as browser navigation (back, forward)
- Added a "selected player" concept to ranking, a ranking that will always show even if outside the page range. Will show as first row of table if rank is above what is visible on the page, last row if rank is below. Selected player row also has a button to jump to the page the selected player appears on. Currently selected player is just the logged in player, but could easily be expanded to allow selecting a different player. I left this out for now because the ranking page is public and the player search functionality isn't.


UT4MasterServer.Web/src/routes.ts
```diff
@@ -4,7 +4,8 @@ import { adminGuard, privateGuard, publicGuard } from './route-guards';
 export const routes: RouteRecordRaw[] = [
   // Secure Pages
   {
-    path: `/Stats/:accountId?`,
+    name: 'Stats',
+    path: `/Stats/:window?/:accountId?`,
```
timiimit, line 8: I think this should be swapped, so that just having account id in url is possible.
```suggestion
    path: `/Stats/:accountId?/:window?`,
```

timiimit (approved):

PR #194, opened 2023-02-17
Restart mongo

timiimit:
This PR makes mongo container restart if it stops.

This should prevent recent incidents when it stops/crashes for an unknown reason. Because our way of rebooting the server causes us to lose logs we have no way of pinpointing exact cause yet. Now if it will auto-restart, logs of container should stay saved which will provide further insight. The downside of this PR is that there's no way for us to know when container stops, hopefully logs will make that clear when it happens.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -30,6 +30,7 @@ services:
   mongo:
     container_name: mongo
     image: mongo
+    restart: always
     environment:
       - MONGO_INITDB_ROOT_USERNAME=devroot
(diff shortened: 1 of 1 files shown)


belmirp (approved):

timiimit:
Waiting a bit before merging in case issue was resolved with apache and it is no longer necessery to restart.

Saibamen (approved):

timiimit:
This doesn't need to happen. issue is not part of our code. our host machine was running out of memory and stopping mongo as result because it was using most of the memory.

Issue #139, opened 2023-01-25
Add email verification


timiimit:
When user registers, email should be sent to his email address. This email would contain a link which verifies the account once it is clicked. If account is not verified within an hour since account creation account should be deleted.

All existing users would have 1 month of time to verify their accounts before they are deleted as well in case they are not verified through email.

Make sure to ban public email services such as `mailinator.com`.

PR #201, opened 2023-02-20
AWS SES implementation

belmirp:
This PR implements the following:

- Sending email as text or HTML via AWS SES
- Sending activation link upon registering new account
- Preventing users which are not in Active status from logging in
- Resending activation link
- Requesting password reset and sending reset link to the email
- Resetting password


Saibamen:
Please fix warnings from linters

UT4MasterServer.Web/src/services/http.service.ts
```diff
@@ -9,10 +9,10 @@ export interface HttpRequestOptions<T = unknown> {
 }
 
 export class HttpError {
-  code: number;
+  code: string;
```
caseyswilliams, line 12: Shouldn't these always be numbers?

UT4MasterServer.Web/src/pages/ResetPassword.vue
```diff
@@ -0,0 +1,137 @@
...
+    qGuid?.toString() === ''
+  ) {
+    status.value = AsyncStatus.ERROR;
+    errorMessage.value = 'Bad request';
+    return;
+  }
+
+  accountId.value = qAccountId!.toString();
+  guid.value = qGuid!.toString();
+}
```
caseyswilliams, line 115: function parseQueryValues() {
  const { accountId: qAccountId, guid: qGuid } = route.query;
  if(!qAccountId?.length || !qGuid?.length ) {
    status.value = AsyncStatus.ERROR;
    errorMessage.value = 'Bad request';
    return;
  }

  accountId.value = qAccountId as string;
  guid.value = qGuid as string;
}

belmirp: Done.

belmirp:
> Please fix warnings from linters

@Saibamen Where do you see warnings?

UT4MasterServer.Web/src/services/http.service.ts
caseyswilliams: Ah, right. The http status code from the fetch api will be a number though, so we should either change the http service to pull numeric error code or change it so that it makes the http error code a string. Otherwise you'll never be able to correctly compare it because typescript will think it has a string when it has a number.

UT4MasterServer.Web/src/pages/ActivateAccount.vue
```diff
@@ -30,24 +30,21 @@ const accountService = new AccountService();
 async function activateAccount() {
   try {
     status.value = AsyncStatus.BUSY;
-    const { email, guid } = route.query;
+    const { accountId, guid } = route.query;
 
-    if (email?.toString() === '' || guid?.toString() === '') {
+    if (accountId?.toString() === '' || guid?.toString() === '') {
```
caseyswilliams, line 35: You don't need to toString and compare to empty string here, just !accountId?.length will work. The query string vars are either string or string[] or null, so a length of 0 is empty string or empty array. This code actually wouldn't work if they were null or undefined, because undefined !== ''

UT4MasterServer.Web/src/services/http.service.ts
caseyswilliams: I see how you're using it now, I will make some changes to it, don't worry about it

caseyswilliams:
> > Please fix warnings from linters
> 
> @Saibamen Where do you see warnings?

There are warnings about non-null assertion. I commented how they can be fixed.

caseyswilliams:
This is good to merge to the feature branch, I will fix up the commented areas.

UT4MasterServer.Web/src/services/http.service.ts
belmirp: We could use `numericErrorCode` instead, but then you would have "magic" numbers on the frontend. I think string representation on an error is better for understanding.
And when I say `errorCode`, I am not referring to the HTTP response code (400 or similar) which is a number.

UT4MasterServer.Web/src/pages/ActivateAccount.vue
belmirp: At first, it was giving me LocationQueryValue as variable type, that is why I had to do toString, but now when I check, destructuring an object to a new variable actually does converting to string.

**Example**: `/Login?activationLinkSent=done`
**Code**: 
```js
const { activationLinkSent: qActivationLinkSent } = route.query;

  console.log(
    'activationLinkSent:',
    activationLinkSent,
    qActivationLinkSent,
    qActivationLinkSent?.length
  );
```

**Output**: activationLinkSent: RefImpl {__v_isShallow: true, dep: Set(1), __v_isRef: true, _rawValue: false, _value: false} done 4

**Example 2**: `/Login?activationLinkSent=`
**Output**: activationLinkSent: RefImpl {__v_isShallow: true, dep: Set(1), __v_isRef: true, _rawValue: false, _value: false}  0

**Example 3**: `/Login?activationLinkSent`
**Output**: activationLinkSent: RefImpl {__v_isShallow: true, dep: Set(1), __v_isRef: true, _rawValue: false, _value: false} null undefined

I will what kind of check will cover all cases.

UT4MasterServer.Web/src/services/http.service.ts
caseyswilliams: Http status codes are hardly magic numbers, they are well defined, and easily enumerated. And we still have that numeric case, when the error response isn't of type ErrorResponse. I will make it accept either case.

UT4MasterServer.Web/src/pages/ActivateAccount.vue
caseyswilliams: You're getting that RefImpl because you're outputting activationLinkSent rather than activationLinkSent.value.

caseyswilliams: The check should be !activationLinkSent?.length because that is just looking for a falsey value. Undefined is falsey and so is 0.

caseyswilliams: But again, don't worry about it, I will fix it in the feature branch, you can merge this and I will pull and fix, it is easier to show you than explain in comments

PR #205, opened 2023-02-22
Added CleanupService for deleting account and all of its related data

belmirp:
Cleanup service will fetch all accounts that are not in `Active` status and have `LastLoginAt` older than a month, delete them and all of their related data.

### DIFF
--- a/UT4MasterServer.Services/Scoped/CleanupService.cs
+++ b/UT4MasterServer.Services/Scoped/CleanupService.cs
@@ -1,0 +1,75 @@
+﻿using Microsoft.Extensions.Logging;
+using UT4MasterServer.Common;
+using UT4MasterServer.Services.Singleton;
+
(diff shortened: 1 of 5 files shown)


UT4MasterServer.Services/Scoped/CleanupService.cs
```diff
@@ -0,0 +1,75 @@
...
+		this.ratingsService = ratingsService;
+	}
+
+	public async Task RemoveNonActivatedAccountsAsync()
+	{
+		var nonActivatedAccountIds = await accountService.GetNonActivatedAccountsAsync();
+		if (!nonActivatedAccountIds.Any())
+		{
+			return;
+		}
```
timiimit, line 48: why do this check? if empty array list is returned, then none will be removed by `RemoveAccountAndAssociatedDataAsync`.

timiimit (approved):

PR #207, opened 2023-02-22
Email domain checking

belmirp:
Invalidating temporary email domains.


timiimit (approved):

timiimit:
There should probably be some check done on existing accounts at some point

PR #210, opened 2023-02-26
Fix multiple matchmaking requests

caseyswilliams:
- Removed polling from footer
- Check for if a request is already in progress in ServerStore.fetchAllServers


timiimit (approved):
didnt look closely...

PR #229, opened 2024-04-26
Update dependencies

timiimit:
Same as PR #220 but more up to date.
- Updates dependencies
- Replaces deprecated `Microsoft.AspNetCore.Authentication` with `Microsoft.AspNetCore.Identity.UI`

### DIFF
--- a/UT4MasterServer.Authentication/UT4MasterServer.Authentication.csproj
+++ b/UT4MasterServer.Authentication/UT4MasterServer.Authentication.csproj
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.Authorization" Version="8.0.4" />
+    <PackageReference Include="Microsoft.AspNetCore.Authorization" Version="6.0.29" />
     <PackageReference Include="Microsoft.AspNetCore.Identity.UI" Version="6.0.29" />
   </ItemGroup>
 
--- a/UT4MasterServer.Authentication/UT4MasterServer.Authentication.csproj
+++ b/UT4MasterServer.Authentication/UT4MasterServer.Authentication.csproj
@@ -7,8 +7,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.Authentication" Version="2.2.0" />
-    <PackageReference Include="Microsoft.AspNetCore.Authorization" Version="6.0.13" />
+    <PackageReference Include="Microsoft.AspNetCore.Authorization" Version="8.0.4" />
+    <PackageReference Include="Microsoft.AspNetCore.Identity.UI" Version="6.0.29" />
   </ItemGroup>
 
   <ItemGroup>
--- a/UT4MasterServer.Services/UT4MasterServer.Services.csproj
+++ b/UT4MasterServer.Services/UT4MasterServer.Services.csproj
@@ -7,8 +7,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="7.0.0" />
-    <PackageReference Include="Microsoft.Extensions.Options" Version="7.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="8.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Options" Version="8.0.2" />
   </ItemGroup>
 
   <ItemGroup>
--- a/UT4MasterServer.Services/UT4MasterServer.Services.csproj
+++ b/UT4MasterServer.Services/UT4MasterServer.Services.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFramework>net6.0</TargetFramework>
@@ -7,8 +7,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="8.0.0" />
-    <PackageReference Include="Microsoft.Extensions.Options" Version="8.0.2" />
+    <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="6.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Options" Version="6.0.0" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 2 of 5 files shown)


UT4MasterServer.Authentication/UT4MasterServer.Authentication.csproj
```diff
@@ -7,8 +7,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.Authentication" Version="2.2.0" />
-    <PackageReference Include="Microsoft.AspNetCore.Authorization" Version="6.0.13" />
+    <PackageReference Include="Microsoft.AspNetCore.Authorization" Version="8.0.4" />
+    <PackageReference Include="Microsoft.AspNetCore.Identity.UI" Version="6.0.29" />
```
Saibamen, line 11: Why are you mixing Microsoft.AspNetCore.* version dependencies?

timiimit: Because I havent worked with .NET in a while and NuGet is telling me to update packages to 8.*.*. Newertheless I'll change it to highest 6.*.* version.

UT4MasterServer.Services/UT4MasterServer.Services.csproj
```diff
@@ -7,8 +7,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="7.0.0" />
-    <PackageReference Include="Microsoft.Extensions.Options" Version="7.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="8.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Options" Version="8.0.2" />
```
Saibamen, line 11: Here too. Please follow package named from docs: https://learn.microsoft.com/en-us/aspnet/core/migration/70-80?view=aspnetcore-8.0&tabs=visual-studio#update-package-references

Saibamen (changes requested):

UT4MasterServer.Services/UT4MasterServer.Services.csproj
timiimit: This is confusing.

`Microsoft.Extensions.*` packages on latest versions (*8.\*.\**) have `net6.0` listed in dependencies.
`Microsoft.AspNetCore.*` packages on latest versions (*8.\*.\**) have only `net8.0`.

Is it preferred to use *8.\*.\** versions even when targeting .NET 6?
Were you in previous comment expecting me to change all packages to absolute latest versions or to *6.\*.\**?

timiimit: Even though `Microsoft.AspNetCore.*` has only `net8.0` as dependency NuGet Manager allows to update to it. Should I just let manager update whatever it can to latest version?

Saibamen: >Even though Microsoft.AspNetCore.* has only net8.0 as dependency NuGet Manager allows to update to it.

It will allow, but installation should fail with error about wrong targetframework

I'm still looking for one answer from Microsoft about these packages. IIRC he said that updating packages 

But you are right -> Extensions are more independent. For example: https://www.nuget.org/packages/Microsoft.Extensions.Options/#dependencies-body-tab

Saibamen: https://www.reddit.com/r/csharp/comments/l6se45/should_i_update_my_microsoft_packages_to_version/

Saibamen: Found it: https://github.com/dotnet/extensions/issues/2700#issuecomment-580413093

timiimit: > It will allow, but installation should fail with error about wrong targetframework

You say that it should fail, but it didn't for me. I was able to successfully build and run whole solution.

> Found it: [dotnet/extensions#2700 (comment)](https://github.com/dotnet/extensions/issues/2700#issuecomment-580413093)

The page you linked answers only how to handle `Microsoft.Extensions.*` packages. It says they need to have the same major version, and we do. All `Microsoft.Extensions.*` have major version 8.

Only author of the question mentioned that you should generally follow semantic versioning based on targetframework, but source of this information is unknown and therefore this information is not trustworthy.

I am just wondering if package major versions are in any way related to targetframework? I'm starting to think that they are not because I am able to build when most packages are on latest version that nuget allows to update to even though all our projects target `net6.0`.

But just in case, yes, lets stick to major version 6 on all `Microsoft.*` packages.

timiimit: This change in `Microsoft.Extensions.*` packages will essentially be a downgrade because we were using major version 7 before.

UT4MasterServer/UT4MasterServer.csproj
```diff
@@ -11,12 +11,12 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.VisualStudio.Azure.Containers.Tools.Targets" Version="1.17.0" />
-    <PackageReference Include="Microsoft.VisualStudio.Web.CodeGeneration.Design" Version="6.0.11" />
+    <PackageReference Include="Microsoft.VisualStudio.Azure.Containers.Tools.Targets" Version="1.20.1" />
+    <PackageReference Include="Microsoft.VisualStudio.Web.CodeGeneration.Design" Version="6.0.16" />
     <PackageReference Include="Serilog.Enrichers.Thread" Version="3.1.0" />
     <PackageReference Include="Serilog.Extensions.Logging" Version="3.1.0" />
     <PackageReference Include="Serilog.Settings.Configuration" Version="3.4.0" />
```
Saibamen, line 18:
```suggestion
    <PackageReference Include="Serilog.Extensions.Logging" Version="8.0.0" />
    <PackageReference Include="Serilog.Settings.Configuration" Version="8.0.0" />
```

Changelogs:
https://github.com/serilog/serilog-extensions-logging/releases
https://github.com/serilog/serilog-settings-configuration/releases

Saibamen (changes requested):

UT4MasterServer/UT4MasterServer.csproj
timiimit: I don't think we should upgrade these because they depend on `Microsoft.Extensions.*` version 8.0.0. Which would cause a mismatch in major versions as we previously discussed.

See here: https://www.nuget.org/packages/Serilog.Extensions.Logging/8.0.0?_src=template#dependencies-body-tab

timiimit: We should probably upgrade whole solution to .NET 8 soon to keep up with the c# ecosystem.

Saibamen: OK. I will update our runtime in near future

Saibamen (approved):
LGTM

PR #230, opened 2024-04-26
Allow in-browser view of cloud files

timiimit:
`Content-Type` is now determined based on file's extension instead of always being `application/octet-stream`.
This prevented cloud files from being displayed in the browser.
This change was made so that announcements can be stored in cloud files and then properly displayed in-game.


UT4MasterServer/Controllers/Epic/CloudStorageController.cs
```diff
@@ -35,8 +38,50 @@ public async Task<IActionResult> ListUserFiles(string id)
...
+	public async Task<IActionResult> UpdateUserFile(string id, string filename)
+	{
+		if (User.Identity is not EpicUserIdentity user)
+			return Unauthorized();
+		var accountID = EpicID.FromString(id);
+		if (user.Session.AccountID != accountID)
+		{
+			// cannot modify other's files
+
+			// unless you are a game server with this player and are modifying this player's stats file
```
Saibamen, line 57:
```suggestion
			// cannot modify other's files
			// unless you are a game server with this player and are modifying this player's stats file
```

UT4MasterServer.Services/UT4MasterServer.Services.csproj
```diff
@@ -7,8 +7,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="7.0.0" />
-    <PackageReference Include="Microsoft.Extensions.Options" Version="7.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="8.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Options" Version="8.0.2" />
```
Saibamen, line 11: Please use 6.x

Saibamen (changes requested):

UT4MasterServer.Services/UT4MasterServer.Services.csproj
timiimit: I messed up my commits and included dependencies here by mistake. I will rewrite history and force push.

timiimit:
I should note that I was not able to get server running without updating the dependencies first. I'm not really sure why. So this should probably only be merged after #229.

Saibamen (approved):

PR #234, opened 2024-04-27
Optimize docker build

timiimit:
There was an issue on hosted server where image building would fail when log/db files used a lot of storage (system ran out of space). This fixes that issue by leaving unneeded files (most notably database and its backups) out of UT4MasterServer's intermediate build stages.

Additionally this fixes the issue where docker was unable to restore all c# projects at the `build` stage.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -27,5 +27,8 @@ README.md
 # Our rules
 
 **/.github
-OldReferenceCode/**
-Files/Engine.ini
+./OldReferenceCode
+./Files
+./db*
+**/logs
+**/Logs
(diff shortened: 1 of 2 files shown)


.dockerignore
```diff
@@ -27,5 +27,8 @@ README.md
 # Our rules
 
 **/.github
-OldReferenceCode/**
-Files/Engine.ini
+./OldReferenceCode
+./Files
+./db*
+**/logs
+**/Logs
```
Saibamen, line 34:
```suggestion
**/[Ll]og
**/[Ll]ogs
```

Saibamen (approved):

PR #248, opened 2026-06-16
Smoke stack integration

itpick:


timiimit:
There's so much wrong with this. That I don't know if i should even properly review this...

Random notes as I skim across the PR:
- Email verification has already been implemented in #213 I just haven't had the time to give it a test, because I would need to setup a test server. It should be a separate PR if for some reason you wish to have a different take on its implementation.
- I saw that `Friends popup on Linux` is supposedly not feasible, which is not true, because I did it with UT4UU without source recompile...
- "UT4UU plugin authors (Letgam3rs)" what???
- "The UT4 binary has Epic prod hostnames hardcoded" false, there are configs in ini which can change them
- QuickPlay is not to be part of UT4MS project. QuickPlay servers can be hosted by anyone. UT4MS should only enable the buttons on main menu once QuickPlay is up in some way.

If this really does what it's supposed to then great, but it introduces so much noise that I am not willing to include it in a current state. Every change should be a separate PR to make changes human readable.
I really don't want to be rude, but honestly it looks like a joke PR. Just like all vibecoding.

PR #249, opened 2026-07-24
feat(matchmaking): let trusted servers override advertised address for NAT/tunnels

itpick:
## Problem

UT4 dedicated servers register with the master via `POST /ut/api/matchmaking/session` (`MatchmakingController.CreateGameServer`). The handler unconditionally overwrites the server's declared `serverAddress` with the **source IP of the registration request**:

```csharp
server.ServerAddress = ipClient.ToString();
```

That is correct for a server on a routable public IP. But for a server behind **NAT or a tunnel** — playit.gg, Cloudflare Spectrum, or a home server behind CGNAT — the source IP the master sees is the server's **egress** address, which is *not* the **ingress** address players must connect to. The result is that the server browser lists an address nobody can reach.

## Change

For **trusted** servers only (`trust != Untrusted`, using the existing `TrustedGameServerService`), an operator can opt in to advertising the real public address. Both mechanisms are **config-gated and default OFF**, so upstream behavior is unchanged unless explicitly enabled. Untrusted-server behavior is left exactly as-is.

A new top-level `Trusted` config section (bound via `IOptions<TrustedGameServerSettings>`, matching the existing settings pattern) provides two opt-ins:

1. **`Trusted:AllowDeclaredAddress`** (bool) — when true, honor the `serverAddress` the server declared in its request body, provided it is a valid, non-empty, non-`"0.0.0.0"` IP. The declared value is captured *before* the source-IP overwrite.
2. **`Trusted:AllowDeclaredAddress` — `X-Forwarded-For` fallback** — the UE5.8 UT client shim (`OnlineSessionUT::RegisterServer`) advertises its `ServerAddressOverride` via an **`X-Forwarded-For` request header** rather than in the body, so the body path above misses those servers entirely. Under the *same* `AllowDeclaredAddress` gate, when the trusted server declares no valid body address, fall back to the **left-most `X-Forwarded-For` entry** (validated the same way: non-empty, non-`"0.0.0.0"`, parseable IP). Captured *before* the source-IP overwrite, same as the body path. Logged with reason `xff-fallback`.
3. **`Trusted:AddressOverrides`** (dictionary `sourceIP -> publicIP`) — otherwise, if the request's source IP is a key in this map, advertise the mapped public IP instead.

**Precedence:** explicit body `declaredAddress` → `X-Forwarded-For` fallback → `AddressOverrides` map → the default source-IP behavior.

When an override is applied, it is logged at Info level (server id, from → to, and reason).

### Example config (environment variables)

```
Trusted__AllowDeclaredAddress=true
Trusted__AddressOverrides__23.123.225.250=69.9.179.19
```

## Notes

- No change to the model or wire format; only the value assigned to `ServerAddress` for trusted servers can differ.
- Built with `dotnet publish -c Release` (the repo Dockerfile) — compiles clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/UT4MasterServer.Models/Settings/TrustedGameServerSettings.cs
+++ b/UT4MasterServer.Models/Settings/TrustedGameServerSettings.cs
@@ -1,0 +1,36 @@
+namespace UT4MasterServer.Models.Settings;
+
+/// <summary>
+/// Optional overrides that let a <em>trusted</em> game server advertise a public
+/// address different from the source IP of its registration request.
+/// </summary>
+/// <remarks>
+/// This exists for game servers behind NAT / tunnels (playit.gg,
+/// Cloudflare Spectrum, home servers behind CGNAT): they register from an
+/// egress IP that is not the address players must connect to, so the browser
+/// would otherwise list an unreachable address. Both mechanisms below are
+/// opt-in and default OFF, so upstream behavior is unchanged unless an
+/// operator explicitly enables them for a trusted server.
+///
+/// Bound from the top-level <c>Trusted</c> configuration section
+/// (e.g. env vars <c>Trusted__AllowDeclaredAddress</c> and
+/// <c>Trusted__AddressOverrides__&lt;sourceIP&gt;</c>).
+/// </remarks>
+public sealed class TrustedGameServerSettings
+{
+	/// <summary>
+	/// When true, a trusted server may declare its own <c>serverAddress</c> in
+	/// the registration request body and it will be honored instead of the
+	/// request source IP, provided the declared value is a valid, non-empty,
+	/// non-<c>"0.0.0.0"</c> IP address.
+	/// </summary>
+	public bool AllowDeclaredAddress { get; set; } = false;
+
+	/// <summary>
+	/// Static map of request source IP (egress) to public address (ingress).
+	/// When a trusted server registers from a source IP present as a key here,
+	/// the mapped value is advertised instead. Consulted only when
+	/// <see cref="AllowDeclaredAddress"/> did not already override the address.
+	/// </summary>
+	public Dictionary<string, string> AddressOverrides { get; set; } = new();
+}
--- a/UT4MasterServer/Controllers/UT/MatchmakingController.cs
+++ b/UT4MasterServer/Controllers/UT/MatchmakingController.cs
@@ -31,15 +31,18 @@ public sealed class MatchmakingController : JsonAPIController
 	private readonly TrustedGameServerService trustedGameServerService;
 
 	private readonly IOptions<ApplicationSettings> configuration;
+	private readonly IOptions<TrustedGameServerSettings> trustedServerSettings;
 
 	public MatchmakingController(
 		ILogger<MatchmakingController> logger,
 		IOptions<ApplicationSettings> configuration,
+		IOptions<TrustedGameServerSettings> trustedServerSettings,
 		MatchmakingService matchmakingService,
 		ClientService clientService,
 		TrustedGameServerService trustedGameServerService) : base(logger)
 	{
 		this.configuration = configuration;
+		this.trustedServerSettings = trustedServerSettings;
 		this.matchmakingService = matchmakingService;
 		this.clientService = clientService;
 		this.trustedGameServerService = trustedGameServerService;
@@ -74,6 +77,11 @@ public sealed class MatchmakingController : JsonAPIController
 		server.ID = EpicID.GenerateNew();
 		server.LastUpdated = DateTime.UtcNow;
 
+		// Capture any address the server declared in its request body BEFORE we
+		// overwrite it, so a trusted server behind NAT/tunnels can opt in to
+		// advertising it (see the override block below).
+		var declaredAddress = server.ServerAddress;
+
 		server.ServerAddress = ipClient.ToString();
 		server.Started = false;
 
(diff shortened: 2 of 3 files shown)


itpick:
@timiimit - please merge and update master server with this so that I can connect my hub.  Thanks!

itpick:
@Saibamen  - would be awesome if you can see this one getting merged in!  Thanks :)

itpick:
@timiimit - you alive?

Saibamen (approved):

itpick:
@Saibamen @timiimit  - I can send you guys a sandwich if you get this in!

UT4MasterServer/Program.cs
```diff
@@ -64,6 +64,7 @@ public static void Main(string[] args)
 		builder.Services
 			.Configure<ApplicationSettings>(builder.Configuration.GetSection("ApplicationSettings"))
 			.Configure<StatisticsSettings>(builder.Configuration.GetSection("StatisticsSettings"))
+			.Configure<TrustedGameServerSettings>(builder.Configuration.GetSection("Trusted"))
```
timiimit: Keep formatting the same as in other lines.
```suggestion
			.Configure<TrustedGameServerSettings>(builder.Configuration.GetSection("TrustedSettings"))
```

UT4MasterServer/Controllers/UT/MatchmakingController.cs
```diff
@@ -85,6 +117,64 @@ public async Task<IActionResult> CreateGameServer([FromBody] GameServer server)
...
+		if (trust != GameServerTrust.Untrusted)
+		{
+			var overrides = trustedServerSettings.Value;
+
+			// The override map is keyed by the client's egress IP. Normalize an
+			// IPv4-mapped IPv6 source (e.g. "::ffff:1.2.3.4") back to plain IPv4 so it
+			// matches config keys written as IPv4.
+			var overrideKey = ipClient.IsIPv4MappedToIPv6 ? ipClient.MapToIPv4().ToString() : ipClient.ToString();
+
+			if (overrides.AllowDeclaredAddress
```
timiimit: What are these 3 `if`s? Which human being writes the same conditions like this in every if statement?

UT4MasterServer/Controllers/UT/MatchmakingController.cs
```diff
@@ -85,6 +117,64 @@ public async Task<IActionResult> CreateGameServer([FromBody] GameServer server)
...
+				server.ServerAddress = forwardedForAddress;
+			}
+			// Validate the config-driven override the same way as the declared one:
+			// trim, reject empty/"0.0.0.0", and require a parseable IP so a misconfigured
+			// entry can't make the master advertise an invalid/unreachable address.
+			// Null-guard AddressOverrides (e.g. JSON "AddressOverrides": null) so
+			// TryGetValue can't throw, and look it up by the normalized IPv4 key.
+			else if (overrides.AddressOverrides is { } addressOverrides
+				&& addressOverrides.TryGetValue(overrideKey, out var mappedAddress)
+				&& mappedAddress?.Trim() is { Length: > 0 } trimmedMapped
```
timiimit: Isn't `mappedAddress?.Trim() is { Length: > 0 } trimmedMapped` equivalent to just `string.IsNullOrWhiteSpace(mappedAddress?.Trim())`?

UT4MasterServer/Controllers/UT/MatchmakingController.cs
```diff
@@ -74,6 +77,35 @@ public async Task<IActionResult> CreateGameServer([FromBody] GameServer server)
 		server.ID = EpicID.GenerateNew();
 		server.LastUpdated = DateTime.UtcNow;
 
+		// Capture any address the server declared in its request body BEFORE we
+		// overwrite it, so a trusted server behind NAT/tunnels can opt in to
+		// advertising it (see the override block below).
+		// Trim so a stray-whitespace value (e.g. " 0.0.0.0") cannot slip past the
+		// "0.0.0.0" guard below while IPAddress.TryParse still accepts it.
+		var declaredAddress = server.ServerAddress?.Trim();
```
timiimit: Why is `declaredAddress` computed even when it's not needed?
This should be inside `if (overrides.AllowDeclaredAddress)` in added chunk of code below.

UT4MasterServer.Models/Settings/TrustedGameServerSettings.cs
```diff
@@ -0,0 +1,36 @@
+namespace UT4MasterServer.Models.Settings;
+
+/// <summary>
+/// Optional overrides that let a <em>trusted</em> game server advertise a public
+/// address different from the source IP of its registration request.
+/// </summary>
+/// <remarks>
+/// This exists for game servers behind NAT / tunnels (playit.gg,
+/// Cloudflare Spectrum, home servers behind CGNAT): they register from an
+/// egress IP that is not the address players must connect to, so the browser
```
timiimit: Why is `browser` mentioned?

itpick:
🙏 please !

UT4MasterServer.Models/Settings/TrustedGameServerSettings.cs
```diff
@@ -0,0 +1,36 @@
+namespace UT4MasterServer.Models.Settings;
+
+/// <summary>
+/// Optional overrides that let a <em>trusted</em> game server advertise a public
+/// address different from the source IP of its registration request.
+/// </summary>
+/// <remarks>
+/// This exists for game servers behind NAT / tunnels (playit.gg,
+/// Cloudflare Spectrum, home servers behind CGNAT): they register from an
```
timiimit: There's no reason to list examples in code. It would be better to explain NAT instead, if anything at all.

UT4MasterServer.Models/Settings/TrustedGameServerSettings.cs
```diff
@@ -0,0 +1,36 @@
...
+/// egress IP that is not the address players must connect to, so the browser
+/// would otherwise list an unreachable address. Both mechanisms below are
+/// opt-in and default OFF, so upstream behavior is unchanged unless an
+/// operator explicitly enables them for a trusted server.
+///
+/// Bound from the top-level <c>Trusted</c> configuration section
+/// (e.g. env vars <c>Trusted__AllowDeclaredAddress</c> and
+/// <c>Trusted__AddressOverrides__&lt;sourceIP&gt;</c>).
```
timiimit: What is this part of the comment talking about? There are no env vars in this PR anywhere.

UT4MasterServer/Controllers/UT/MatchmakingController.cs
```diff
@@ -85,6 +117,64 @@ public async Task<IActionResult> CreateGameServer([FromBody] GameServer server)
...
+				&& IPAddress.TryParse(trimmedMapped, out _))
+			{
+				logger.LogInformation(
+					"Trusted server {ServerID} (client {ClientID}) address override {From} -> {To} (reason: egress->ingress map)",
+					server.ID, server.OwningClientID, server.ServerAddress, trimmedMapped);
+				server.ServerAddress = trimmedMapped;
+			}
+		}
+
 		if (trust != GameServerTrust.Untrusted)
```
timiimit: Literally the same if statement is just below yours. Why not put your code in this one?

UT4MasterServer/Controllers/UT/MatchmakingController.cs
```diff
@@ -85,6 +117,64 @@ public async Task<IActionResult> CreateGameServer([FromBody] GameServer server)
...
+			// TryGetValue can't throw, and look it up by the normalized IPv4 key.
+			else if (overrides.AddressOverrides is { } addressOverrides
+				&& addressOverrides.TryGetValue(overrideKey, out var mappedAddress)
+				&& mappedAddress?.Trim() is { Length: > 0 } trimmedMapped
+				&& trimmedMapped != "0.0.0.0"
+				&& IPAddress.TryParse(trimmedMapped, out _))
+			{
+				logger.LogInformation(
+					"Trusted server {ServerID} (client {ClientID}) address override {From} -> {To} (reason: egress->ingress map)",
+					server.ID, server.OwningClientID, server.ServerAddress, trimmedMapped);
```
timiimit: All blocks with your if-statements contain the same duplicated code. Why not add another parameter for "reason" to `LogInformation` instead?

itpick:
don't forget!

